PR #14, opened 2023-06-11
Default port now 4242 everywhere

ferdinandschober:


ferdinandschober:
the port for new client defaulted to a wrong port

PR #42, opened 2023-12-04
macos: enable running lan-mouse on macos

feschber:
- adapted conditional compilation
- forked frontend into its own process
- moved lan-mouse socket to ~/Library/Caches/lan-mouse-socket.sock instead of XDG_RUNTIME_DIR


feschber:
// TODO: Keycode translation

PR #81, opened 2024-01-24
macos: add keyboard support

kiwiz:
`keycode` is doing all the heavy lifting here.

I encountered some odd behaviour with Caps lock (which I've mapped to Control) getting stuck, but keys appear to be translated properly otherwise.

### DIFF
--- a/src/backend/consumer/macos.rs
+++ b/src/backend/consumer/macos.rs
@@ -10,9 +10,15 @@ use core_graphics::event::{
 use core_graphics::event_source::{CGEventSource, CGEventSourceStateID};
 use keycode::{KeyMap, KeyMapping};
 use std::ops::{Index, IndexMut};
+use std::time::Duration;
+use tokio::task::AbortHandle;
+
+const DEFAULT_REPEAT_DELAY: Duration = Duration::from_millis(500);
+const DEFAULT_REPEAT_INTERVAL: Duration = Duration::from_millis(32);
 
 pub struct MacOSConsumer {
     pub event_source: CGEventSource,
+    repeat_task: Option<AbortHandle>,
     button_state: ButtonState,
 }
 
@@ -60,6 +66,7 @@ impl MacOSConsumer {
         Ok(Self {
             event_source,
             button_state,
+            repeat_task: None,
         })
     }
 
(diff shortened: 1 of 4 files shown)


src/backend/consumer/macos.rs
```diff
@@ -209,9 +210,14 @@ impl EventConsumer for MacOSConsumer {
                 PointerEvent::Frame { .. } => {}
             },
             Event::Keyboard(keyboard_event) => match keyboard_event {
-                KeyboardEvent::Key { .. } => {
-                    /*
-                    let code = CGKeyCode::from_le(key as u16);
+                KeyboardEvent::Key { time: _, key, state } => {
+                    let code = match KeyMap::from_key_mapping(KeyMapping::Evdev(key as u16)) {
+                        Ok(k) => CGKeyCode::from_le(k.mac),
```
feschber, line 215: I'm unsure about the endianness here: Should this be `from_ne(k.mac)` maybe? I had that there more or less as a placeholder.

feschber (changes requested):
Awesome work! I didn't know about this crate.

It seems to work for me without issues. The only thing that does not work is repeating keys, it seems.

You could take some inspiration from https://github.com/feschber/lan-mouse/blob/8084b52cfc8c661baa43fbea158e0b44cd0ae203/src/backend/consumer/windows.rs#L91 but that may as well be a new PR.

Rust format is complaining but other than that, lgtm!

src/backend/consumer/macos.rs
kiwiz: `from_ne` doesn't appear to be defined. I tried `from_be` too, which gives odd results (which makes sense since `u16` is 2 bytes).

kiwiz (commented):
Pushed up my attempt to integrate the key repeating logic. I tried turning `key_event` into a method, but it might be simpler to just `event_source` and to pass it into `key_event` in the event handler and the timer?

src/backend/consumer/macos.rs
feschber: Actually there is no conversion necessary at all. I believe I missed that it's just a typedef to u16. So this should be `k.mac as CGKeyCode` then.

kiwiz:
That should do it! I'm new to Rust so feedback welcome.

feschber (approved):

feschber:
> That should do it! I'm new to Rust so feedback welcome.

Thank you, looking good!
 Aren't we all new to Rust?
 
 I fixed one clippy complaint and updated the README, hope you don't mind :)

PR #94, opened 2024-03-16
Libei producer

### LINKED ISSUE #86: Mouse gets stuck in gnome
lan-mouse 0.5.1 on both machines. Primary machine using sway, secondary machine using gnome 45.3.

Pointer can enter gnome and control works just fine. It is however unable to leave gnome by moving the pointer, either from the primary PC's mouse, or the gnome machines' mouse.

feschber:
closes #86

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -97,8 +97,9 @@ dependencies = [
 
 [[package]]
 name = "ashpd"
-version = "0.9.0"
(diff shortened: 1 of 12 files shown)


feschber:
Todos:

- [ ] fix panic on exit
- [ ] distinguish "cancel" and other errors when creating capture session

PR #108, opened 2024-04-24
nix: Add aarch64-darwin package and devshell

meck:
Enable building and developing using nix on aarch64-darwin

It should be possible to just add `x86_64-darwin` to systems as well, however i don't have machine to test it on hand

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -15,6 +15,8 @@
   }: let
     inherit (nixpkgs) lib;
     genSystems = lib.genAttrs [
+      "aarch64-darwin"
+      "x86_64-darwin"
       "x86_64-linux"
     ];
     pkgsFor = system:
@@ -49,6 +51,8 @@
           gtk4
           libadwaita
           xorg.libXtst
+        ] ++ lib.optionals stdenv.isDarwin [
+          darwin.apple_sdk_11_0.frameworks.CoreGraphics
         ];
 
         RUST_SRC_PATH = "${rust}/lib/rustlib/src/rust/library";
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -18,6 +18,8 @@ rustPlatform.buildRustPackage {
     gtk4
     libadwaita
     xorg.libXtst
+  ] ++ lib.optionals stdenv.isDarwin [
+    darwin.apple_sdk_11_0.frameworks.CoreGraphics
   ];
 
   src = builtins.path {
--- a/.github/workflows/cachix.yml
+++ b/.github/workflows/cachix.yml
@@ -3,8 +3,14 @@ name: Binary Cache
 on: [push, pull_request, workflow_dispatch]
 jobs:
   nix:
+    strategy:
+      matrix:
+       os: 
+        - ubuntu-latest
+        - macos-latest
+        - macos-14
     name: "Build"
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }} 
     steps:
     - name: Checkout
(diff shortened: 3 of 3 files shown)


feschber:
> It should be possible to just add `x86_64-darwin` to systems as well, however i don't have machine to test it on hand

Could you update the CI (https://github.com/feschber/lan-mouse/blob/main/.github/workflows/cachix.yml) and add a nix-macos job that runs on macos-14?

For x86_64 you could do the same (macos-latest is the runner for x86_64)

flake.nix
```diff
@@ -45,10 +46,13 @@
           rust
           rust-analyzer-unwrapped
           pkg-config
-          xorg.libX11
           gtk4
           libadwaita
+        ] ++ lib.optionals stdenv.isLinux [
+          xorg.libX11
```
feschber, line 52: I would like to have this in a separate PR, as it is not directly related to this one

nix/default.nix
```diff
@@ -14,10 +14,13 @@ rustPlatform.buildRustPackage {
   ];
 
   buildInputs = with pkgs; [
-    xorg.libX11
     gtk4
     libadwaita
+  ] ++ lib.optionals stdenv.isLinux [
+    xorg.libX11
```
feschber, line 20: same thing here

meck:
> > It should be possible to just add `x86_64-darwin` to systems as well, however i don't have machine to test it on hand
> 
> Could you update the CI (https://github.com/feschber/lan-mouse/blob/main/.github/workflows/cachix.yml) and add a nix-macos job that runs on macos-14?
> 
> For x86_64 you could do the same (macos-latest is the runner for x86_64)
There seems to be some issue with the x86 build in CI, Ill see if i can figure it out.

meck:
Turns out nixpgks uses different apple SDKs for building darwin aarch64/x86, by specifying the specific one it builds and produces an x86 executable that runs fine (under emulation on a aarch64 machine)

meck:
Thanks, I did a quick check and running `nix build nixpkgs#lan-mouse  --option system x86_64-darwin` on an m1 mac seems to fail with same issue (aarch64 works fine). Here is the [note about the SDK](https://github.com/NixOS/nixpkgs/blob/6b78a949091c51385c46c7e8e7ce9293b5098fc8/doc/stdenv/platform-notes.chapter.md?plain=1#L64)

feschber:
> Thanks, I did a quick check and running `nix build nixpkgs#lan-mouse --option system x86_64-darwin` on an m1 mac seems to fail with same issue (aarch64 works fine). Here is the [note about the SDK](https://github.com/NixOS/nixpkgs/blob/6b78a949091c51385c46c7e8e7ce9293b5098fc8/doc/stdenv/platform-notes.chapter.md?plain=1#L64)

So it works when compiling on x86 but not when crosscompiling from aarch64?

meck:
> So it works when compiling on x86 but not when crosscompiling from aarch64?

I don't have an x86 mac but i think the issue would still persist there, as it is related to which SDK is used by nix (separate from the mac host SDK). And i think the x86 build on M1 isn't actually cross-compiling, its running the native compiler under emulation. But im not sure

feschber:
> > So it works when compiling on x86 but not when crosscompiling from aarch64?
> 
> I don't have an x86 mac but i think the issue would still persist there, as it is related to which SDK is used by nix (separate from the mac host SDK). And i think the x86 build on M1 isn't actually cross-compiling, its running the native compiler under emulation. But im not sure

I will check on my VM later.

PR #109, opened 2024-04-24
nix: enable creating config file via home-manager

meck:
In order to have a completely declarative setup using nix


meck:
Yes, i think it should be possible to take in a toml file and convert to nix syntax, i assume you mean `./config.toml`. How about the `./nix/README.md`? could make that an explanation that the option is the same but in nix syntax.

meck:
I've updated the PR to use the config in the repo

feschber:
> I've updated the PR to use the config in the repo

Thank you!

Issue #49, opened 2023-12-15
[Feature Request] Automation


MulverineX:
Ideally there would be a built-in way to setup lan-mouse to run on start up, for all platforms. Currently, especially on Windows, this is a pain in the ass to do.

Some sort of auto-activate flag in the config, and a couple scripts that add the program to startup would be great.

feschber:
It is now possible to activate clients on startup: 0017dbc634abb0a53ec9478f0c1605d7a7805be3

feschber:
A linux systemd user service is in the works, I will see if I can come up with a windows service definition (not done that before)

MulverineX:
https://users.rust-lang.org/t/running-rust-app-without-showing-the-terminal/41859/2

It'd be nice if you would enable this.

```rs
#![windows_subsystem = "windows"]
```

@feschber

PR #110, opened 2024-04-25
nix: Add macOS launchd service to hm-module module

meck:
Home manager has the possibility to run macOS launchd services. These are automatically disabled when running on linux.

### DIFF
--- a/nix/hm-module.nix
+++ b/nix/hm-module.nix
@@ -23,7 +23,12 @@ in {
     systemd = mkOption {
       type = types.bool;
       default = pkgs.stdenv.isLinux;
-      description = "Whether to enable to systemd service for lan-mouse.";
+      description = "Whether to enable to systemd service for lan-mouse on linux.";
+    };
+    launchd = mkOption {
+      type = types.bool;
+      default = pkgs.stdenv.isDarwin;
+      description = "Whether to enable to launchd service for lan-mouse on macOS.";
     };
   };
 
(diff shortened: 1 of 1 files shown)


feschber:
Would be a good addition to have a launchd service definition in general. (#49)

Same as with the example config, I would prefer to have this in a central location (e.g. `service/macos-launchd/lan-mouse.xml`) and then referenced from the nix config.

I'm not to familiar with macos, so let me know if this could work.

meck:
Not quite sure, I'm not really a mac dev either. Pulling an externally defined launchd unit to nix would be complicated as it would need some editing wrt paths and so on. It's basically the same situation as the systemd unit that currently is defined in `hm-module.nix` separate from the one in `service/lan-mouse.service`

Regarding providing a general `launchd` unit outside of nix as a plain file there are some issues as well, the path to the executable needs to be absolute, so might need some editing. The canonical "mac" way would be to crate an app bundle (maybe (cargo-bundle)[https://github.com/burtonageo/cargo-bundle] could be used) and include the unit there and install at runtime.

feschber:
> Not quite sure, I'm not really a mac dev either. Pulling an externally defined launchd unit to nix would be complicated as it would need some editing wrt paths and so on. It's basically the same situation as the systemd unit that currently is defined in `hm-module.nix` separate from the one in `service/lan-mouse.service`
> 
> Regarding providing a general `launchd` unit outside of nix as a plain file there are some issues as well, the path to the executable needs to be absolute, so might need some editing. The canonical "mac" way would be to crate an app bundle (maybe (cargo-bundle)[https://github.com/burtonageo/cargo-bundle] could be used) and include the unit there and install at runtime.  

Okay I see this is something that will go on the list then.

Issue #107, opened 2024-04-21
Invoke custom script when input is switched


xpufx:
Barrier has this parameter `--screen-change-script <path>` : `full path to script to run on screen change first argument is the new screen name` which runs a user defined script when input switching occurs.

I am using this feature to switch my monitor input with ddc commands as detailed here: https://xpufx.com/posts/hundred-percent-software-kvm-switch/

Is there similar functionality in lan-mouse?

feschber:
Also: Are you aware of input-leap? It's the maintained barrier fork.

feschber:
Will need to update for windows and update the documentation

xpufx:
> Also: Are you aware of input-leap? It's the maintained barrier fork.

Somehow I missed this comment. At the time I posted the question none of the barrier forks had wayland support. Now input leap and the new synergy upstream Deskflow do seem to have it.

 Thank you for the enter hook support. Awesome. I will test it out asap.

Issue #105, opened 2024-04-13
Suggestion: Clipboard


DrYak:
For sharing clipboard, I've found something that can both be used as a building block for users to make their workaround and which could be used to add clipboard support into lan-mouse:

[wl-clipboard](https://github.com/bugaevc/wl-clipboard)

It provides a pair of command that can read from- / write to- the wayland clipboard (and thus be used to store stuff in files, send over pipes or tunnels, etc.)

At the end of the readme, they also mention a rust crate [wl-clipboard-rs](https://github.com/YaLTeR/wl-clipboard-rs) handling the same which could be used for lan-mouse.

feschber:
wl-clipboard-rs is definitely a good starting point. It uses the same underlying library as the layer-shell backend in lan-mouse.

There are however some additional problems to solve other than just interacting with the clipboard.

I will just lay out some of my thoughts here for future reference:

1. **_How_** to send the clipboard contents to the destination client reliably? (especially larger clipboard contents)
(a) Use an additional TCP server socket on each client
(b) Chunk the contents into smaller chunks and send them via the same UDP socket.

Option (a) would be a lot easier to implement but I'm definitely leaning towards the second option because it does not require an additional port and protocol. There are however a few issues with it that need solving:
- The clipboard contents must be chunked into a size that fits within UDP packets and later reassembled on the target client.
- Mouse and Keyboard events must not be blocked while clipboard contents are being sent (The mouse should not hang while a 1GiB file is being sent) so this is probably going to require some form of a priority queue (or maybe just a QoS header to let the kernel handle it?).
- Both of the above require changes to the protocol.

2. **_When_** to send the clipboard contents?
(a) When something is copied on the source client
(b) When the cusor enters the destination client
(c) When something is pasted on the destination client

I'm also leaning towards the last option here because it would require the least amount of data being sent but I'm also not sure if it is possible to do via standard wayland protocols so I may just resort to option (b) at first.

DrYak:
Copying the clipboard upon cursor entering client seems indeed the most reasonable.

(And sounds that you almost need to manually implement your own TCP-like layer on top of UDP)

DrYak:
Hum, maybe #107 could be a temporary solution for the clipboard:
just handle it externally of lan-mouse for now (e.g. by using wl-clipboard and the like in external scripts that move the clipboard content around upon moving the pointer to another machine) and so no need to think how to organise transfer of large blocks of data over a piece of software optimised for small and frequent UDP packets.

[...]

PR #130, opened 2024-05-11
basic enter hook command

feschber:
#105 #107


feschber:
does not work on windows yet, because sh is hardcoded.

PR #136, opened 2024-05-22
Extract package name and version from Cargo.toml

ICE-GB:
Extract package name and version from Cargo.toml

- Use builtins.readFile to read Cargo.toml contents.
- Parse TOML content with builtins.fromTOML to extract `package.name` and `package.version`.
- Apply extracted values to set pname and version in the Nix build derivation.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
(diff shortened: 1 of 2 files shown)


feschber:
Thank you, I totally forgot to update the version... Sorry about that

PR #145, opened 2024-06-19
Update Arch Linux instructions

orhun:
`lan-mouse` is now in the official repositories: <https://archlinux.org/packages/extra/x86_64/lan-mouse/> 🥳


feschber:
I feel honored! Thank you!

Issue #159, opened 2024-07-12
Icons missing and emulated pointer invisible in Windows build


vilhalmer:
When capturing input from Wayland and sending to Windows, cursor movement and keystrokes are working just fine. However, the pointer is invisible on the Windows machine. The icons in the UI are also broken images. I'm not sure how input emulation on Windows works, if lan-mouse is supposed to be drawing its own cursor image perhaps these two things are related?

I've tried both the latest release (0.8.0) and the pre-release build from CI.

feschber:
The missing cursor happens, when there is no actual input device plugged into the windows machine. I have not yet figured out how to prevent that.

Missing icons are because the package is not yet properly bundled for windows.
It's a bit better if you build from source but not perfect.

vilhalmer:
Aha, indeed I don't have the mouse connected at the same time I'm doing this. I'll plug one in and drop that part of the issue.

feschber:
I had another crack at this: Seems like installing adwaita-icon-theme fixes this issue.

```sh
gvsbuild build adwaita-icon-theme
```

I'll have to see, how to distribute this correctly

PR #161, opened 2024-07-16
Update Nix Flake

hannesschulze:
- Add support for `aarch64-linux`
- Add `librsvg` to enable support for SVG icons in GTK


feschber:
#159 could be the same issue

Issue #164, opened 2024-07-26
Feature Request: Support for one-way control


b0o:
Really awesome project! One thing that would be nice is that I want to control my Mac from my Linux computer, but not the other way around. When I am focused on my Linux computer, I don't want my Mac's trackpad to be routed to my Linux computer. I always want the trackpad to control my Mac.

Maybe something like this in the config:

```toml
[right]
hostname = "my_mac"
activate_on_startup = true
client_only = true
```

(Note: I am currently using #131 on my Mac)

feschber:
This is a good idea!
 I have already been considering this, since it would open up a few more usecases like multiple devices controlling one other device.
My plan would be to have a separation between "incoming" and "outgoing" connections.

Dregu:
My use case for one-way control would be that the laptop I'm using to remote control my desktop is simply always behind my chair on the couch, so being able to accidentally remote to a laptop I can't see is just detrimental.

That said, what I would actually like to do is keep two-way control as an option for special occasions (when it's actually on my desk), but simply disable any kind of mouse-on-screen-edge detection and instead use keyboard shortcut to connect to a specific host (or release if already connected).

This really goes hand in hand with the other two suggestions, but after disabling screen-edge and adding keyboard shortcuts, hosts should be able to just exist in the network with a name like [steve] instead of [right] or [left].

Currently I am launching lan-mouse with a keyboard shortcut on demand on the laptop when I feel like poking something on the desktop/TV, so that I wouldn't accidentally remote to the laptop, but there's still the extra step of dragging the mouse to the edge.

So yeah, just some thoughts how having these three options instead of a simple one-way switch would be a lot more versatile, while also achieving one-way functionality if configured that way.

feschber:
So this is not fully what you are describing *yet*, however the first and most difficult step is there: Device A controlling device B is now independent of device B controlling device A.

What you are describing is somewhat the same as #26.

So my plan now is to

a) make automatic cursor release on the remote device optional (this would also be a workaround for #132)
b) implement an option to add a keybind for toggling between devices

Removing the capture activation barriers on the sending side is somewhat more difficult because of the way libei works but it might be doable as well.

Issue #104, opened 2024-04-13
Encryption


DrYak:
Hi, I switched to this tool as Plasma switched to Wayland and love it.

A few suggestions regarding encryption.

## inside lan-mouse

I've noticed that [WebRTC-rs](https://github.com/webrtc-rs/webrtc/) have a working implementation of DTLS which would be the cannonical way for input-leap to encrypt UDP streams.

Have you had a look at it? (I am saddly too much of a rust newbie to be of any use making a PR, and even more so with delicate subjects like encryption)

## workarounds outside lan-mouse

Until there's proper encryption inside lan-mouse, a few idea for users to hack their own workarounds:

- I've noticed [dtlspipe](https://github.com/SenseUnit/dtlspipe) which is basically like stunnel but for UDP over DTLS instead of TCP.
- There is also the classic approach of using netcat to forward UDP to TCP connection and use SSH forwarding to establish a secure connection. This sacrifice the "spontaneous connect/disconnect" that lan-mouse's UDP offers, but I think SSH tunnels and the port not even open on the laptops' firewalls is about the most secure solution (unless one has the wrong version of XZ :grin:  )

feschber:
WebRTC dtls is in fact exactly what I was planning to use for this.
I will need to rewrite some of the network code in order to use it but I will hopefully get to it in the coming weeks.

DrYak:
>  but I will hopefully get to it in the coming weeks.

That would be extremely cool! I think encryption (or any user-made hack for security) is the last major blocker before more people can start using this as a Synergy replacement (or depending on interests, until Input Leap and/or Waynergy get the bits they need upstreamed into Kwin, portals, etc.)

BTW: I tried [packaging your software](https://build.opensuse.org/package/show/home:dr-yak/lan-mouse) for my openSuse Tumbleweed. The x86_64 package builds successfully (The i586  variant fails on libadwaita). So I'll be able to easily install accross my machines.

feschber:
> BTW: I tried [packaging your software](https://build.opensuse.org/package/show/home:dr-yak/lan-mouse) for my openSuse Tumbleweed. The x86_64 package builds successfully (The i586 variant fails on libadwaita). So I'll be able to easily install accross my machines.

I'm not very familiar with OpenSuse but free to update the [installation instructions](https://github.com/feschber/lan-mouse#installation) if you want that to be included. (you can make a PR).

If Libadwaita does not want to build on i586, you could deactivate the feature for building (cli should still be usable):

[...]

PR #194, opened 2024-09-10
Add link to Synergy (open source)

nbolton:
Hey @feschber, I admire and respect the work you're doing with Lan Mouse. It's a great project. As a nod to the potential collaboration between our projects, perhaps you could mention [Synergy 1 Community Edition](https://github.com/symless/synergy) (the original ancestor of the Input Leap fork) which is still in development (as it's the open source core of Synergy 3). I am happy to reciprocate a link to Lan Mouse in the Synergy readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ Focus lies on performance and a clean, manageable implementation that can easily
 
 ***blazingly fast™*** because it's written in rust.
 
-For an alternative (with slightly different goals) you may check out [Input Leap](https://github.com/input-leap).
+For an alternative (with slightly different goals) you may check out [Synergy 1 Community Edition](https://github.com/symless/synergy) or [Input Leap](https://github.com/input-leap) (Synergy fork).
 
 
 > [!WARNING]


feschber:
Hey @nbolton! I feel honored to get recognition by the pioneers in this area ;). Will do

nbolton:
Sounds like the start of something great. I'll add a link to Lan Mouse in the Synergy readme. For context, I'm thinking of explaining that the projects have different goals much like you do on your readme, and I believe that it'd be good to explore what those are so I don't misrepresent them. If you're open to it, how would be best for us to discuss this? We can discuss it here or somewhere else if you prefer.

feschber:
Thank you I appreciate the advertisement!

We can discuss here or you can message me via the email address in my profile if you prefer.

In the short term the main goal is of course encryption (#104) and support for single direction control (#164) which I'm currently working on.

Aside from that I'm planning to improve platform support and eventually support Android as well.
 
I think it would be cool to have interop with synergy / barrier as well but I have not yet looked at the protocol and how easy that would be to integrate with the existing architecture.
 
The app is also architected in a way that allows to add native frontends for QT / KDE, Windows and MacOS in the future which is something I wanted to approach once the more pressing issues are resolved.

I also want to work on some testing and benchmarking tools eventually.

nbolton:
Sounds good, I'm looking forward to seeing the testing and benchmarking tools, as this is something I've wanted to do with Synergy for a long time but had to prioritise other things.

We have a [very long-term roadmap/vision](https://symless.com/synergy/roadmap) if that's of interest to you. A lot of it will need to be in Synergy 1 (open source) for the proprietary version to work, so you'll be able to see the source when we eventually come to implement those features. Long term, we only plan on having the GUI as the proprietary code (but things can change and who knows, maybe we'll open source everything).

I [added a link to Lan Mouse](https://github.com/symless/synergy/commit/c9eebe0d1a387461dcbbba9f8de1a4403bc1f9e5), but I don't feel like I did what you said justice. I tried to keep it short and sweet as there is way too much text on the README right now but feel free to submit a PR if you want to tweak the wording, as I know that goals can change over time.

PR #195, opened 2024-09-10
encryption - add ring dependency

jonstelly:
missing dependency in encryption branch #104

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1728,6 +1728,7 @@ dependencies = [
  "local-channel",
  "log",
  "rcgen",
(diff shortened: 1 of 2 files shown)


feschber:
Hi @jonstelly, I appreciate the effort but I'm not accepting PRs on feature branches.

Also it should not be necessary to explicitly state the dependency, as [rustls already depends on ring](https://github.com/rustls/rustls/blob/9973f08ffe0fcae44532352a8970abe6cede3224/rustls/Cargo.toml#L42), when the feature is enabled.

PR #211, opened 2024-10-05
Add default.nix file to main and update readme

emileakbarzadeh:
Adding a default.nix file to the base project so it can be built with nix-build for easy building. Also updated readme for documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,13 @@ paru -S lan-mouse-bin
 - nixpkgs: [search.nixos.org](https://search.nixos.org/packages?channel=unstable&show=lan-mouse&from=0&size=50&sort=relevance&type=packages&query=lan-mouse)
 - flake: [README.md](./nix/README.md)
 
+### Building from source with Nix
+In the root of the project to build Lan Mouse, run
+```sh
+nix-build
+```
+
+You can find the executable in `result/bin/lan-mouse` .
 
 ### Manual Installation
 
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,3 @@
+{ pkgs ? import <nixpkgs> { }
+}:
+pkgs.callPackage nix/default.nix { }


feschber:
Could someone more knowledgeable in nix comment on this?
Does it have anything to do with [nix/default.nix](https://github.com/feschber/lan-mouse/blob/main/nix/default.nix) 
@meck @ICE-GB @hannesschulze ?

hannesschulze:
This is change useful when building the package with the non-flake [`nix-build`](https://nix.dev/manual/nix/2.24/command-ref/nix-build.html) tool. When using flakes, it is already possible to simply build the package using [`nix build`](https://nix.dev/manual/nix/2.24/command-ref/new-cli/nix3-build) from the root of the repository. This is the approach currently supported by the repo; it ensures a reproducible build environment because the dependency on nixpkgs is managed explicitly and pinned in `flake.lock`. While I would personally prefer mentioning this method in the README instead, I do understand there is a bit of a controversy around flakes and officially they are still considered "experimental" even though a lot of people (including me) use them.

feschber:
Considering the nix directory only contains two files, I guess it would make sense to just move the contents to the root of the repository? (flake.nix is there already)

ICE-GB:
I have hardly ever used non-flake. 
However, the default.nix of this pr should be different from the current nix/default.nix file.
flake.nix is ​​the entry point built using the flake method. 
default.nix in this pr is the entry point built using the non-flake method.

feschber:
> I have hardly ever used non-flake. However, the default.nix of this pr should be different from the current nix/default.nix file. flake.nix is ​​the entry point built using the flake method. default.nix in this pr is the entry point built using the non-flake method.

I think I understand now: `flake.nix` refers to `nix/default.nix` with the `callPackage ./nix {};` line and `default.nix` can refer to the same `nix/default.nix`, building it the traditional way instead. I was just confused by `default.nix` and `nix/default.nix` having the same name here.

I have no problem supporting the normal `nix-build` command, just wanted to understand whats going on here!

I would however prefer, if you could merge the sections in the README instead of making it a new section.

Like

### Nix
- Nixpkgs
- Flake
- Build from source

or something along the lines

ICE-GB:
When the file name is default.nix, you can just write the full path to the directory instead of the full path to the file.

PR #241, opened 2024-12-01
fix windows input capture not releasing keys

AdamJedl:


feschber:
Thank you, this is definitely an oversight ;)

Issue #122, opened 2024-05-05
[Help] I can't use him


mystery4f:
I can't use him! My two windows connect and then nothing happens

mystery4f:
(lan-mouse.exe:11616): GLib-GIO-WARNING **: 21:04:35.522: win32 session dbus binary not found
[2024-05-05T13:04:36Z DEBUG lan_mouse::frontend::gtk] connecting to lan-mouse-socket
[2024-05-05T13:04:36Z DEBUG lan_mouse::server::frontend_task] frontend: Enumerate
[2024-05-05T13:04:36Z DEBUG lan_mouse::frontend] json: {"Enumerate":[]}, len: 16
[2024-05-05T13:04:36Z DEBUG lan_mouse::frontend::gtk] connected to lan-mouse-socket
[2024-05-05T13:04:40Z DEBUG lan_mouse::frontend::gtk::window] requesting {"AddClient":[null,4242,"Left"]}
[2024-05-05T13:04:40Z DEBUG lan_mouse::server::frontend_task] frontend: AddClient(None, 4242, Left)
[2024-05-05T13:04:40Z INFO  lan_mouse::server::frontend_task] adding client [left] @ {}
[2024-05-05T13:04:40Z DEBUG lan_mouse::server::frontend_task] add_client 0
[2024-05-05T13:04:40Z DEBUG lan_mouse::frontend] json: {"NotifyClientCreate":{"hostname":null,"fix_ips":[],"handle":0,"ips":[],"port":4242,"pos":"Left"}}, len: 98
[2024-05-05T13:04:41Z DEBUG lan_mouse::frontend::gtk::client_row::imp] state change -> requesting update
[2024-05-05T13:04:41Z DEBUG lan_mouse::frontend::gtk::window] requesting update: UpdateClient(0, None, 4242, Left)
[2024-05-05T13:04:41Z DEBUG lan_mouse::frontend::gtk::window] requesting {"UpdateClient":[0,null,4242,"Left"]}
[2024-05-05T13:04:41Z DEBUG lan_mouse::frontend::gtk::window] requesting activate: ActivateClient(0, true)
[2024-05-05T13:04:41Z DEBUG lan_mouse::server::frontend_task] frontend: UpdateClient(0, None, 4242, Left)
[2024-05-05T13:04:41Z DEBUG lan_mouse::frontend::gtk::window] requesting {"ActivateClient":[0,true]}
[...]

PR #246, opened 2025-01-03
 #122 chore: added additional using information

Truenya:
several times, trying to master this new tool for myself, I came across one problem. Fortunately, #122 helped. I think this deserves a place in the readme


feschber:
I'm going to close this, as it does not apply anymore.
The earlier version of the README stated that both devices had to be connected and activated.
This definitely was not very intuitive which is why it is now changed.

PR #270, opened 2025-02-21
Add cmd-q shortcut on macOS

mrled:
Wire up the Quit action to the File menu on macOS.

This enables the typical Cmd-Q shortcut for quitting the app, which is a standard on macOS. To do this, I created a global menu, which macOS uses automatically. (Actually, there is always a menu bar on macOS no matter what. This change enables the Quit menu item, which exists but is grayed out otherwise.) The menu doesn't appear on other platforms.

If this is a wrongheaded way to do this, I'd be happy to change it, and not offended if you just need to close.


feschber:
Looks good to me ;)

PR #280, opened 2025-03-17
Add a Windows msi installer

feschber:


feschber:
Todos
[ ] include the gtk dlls
[ ] check if any other dlls are requried

nbolton:
Do Rust binaries depend on MSVC? Just curious. I'm pretty sure Rust uses MSVC Toolchain by default.

If it does, you may want to add an MSVC dependency to the WiX config. Here's how I did it recently: https://github.com/deskflow/deskflow/blob/master/deploy/windows/wix-patch.xml.in

You can also use WiX `burn` to bundle the MSVC redist MSI with your MSI to create a .exe that installs both (this is the recommended approach since Microsoft has deprecated merge modules which were historically used to install the MSVC deps).

nbolton:
> Is there any proper documentation on how to do this?

Unfortunately, I did not, just had to muddle my way through it, sadly. For WiX, `docs.firegiant.com` seems to be the best but Microsoft docs for creating MSI files seems hard to discover. I can't remember where I read about side-loading and merge modules being discouraged and why... but this seems firmly engrained in my memory. Microsoft really wants us to redistribute their MSI redistributable... I think it's easier for them to keep up to date and patch security vulnerabilities.

feschber:
> > Is there any proper documentation on how to do this?
> 
> Unfortunately, I did not, just had to muddle my way through it, sadly. For WiX, `docs.firegiant.com` seems to be the best but Microsoft docs for creating MSI files seems hard to discover. I can't remember where I read about side-loading and merge modules being discouraged and why... but this seems firmly ingrained in my memory. Microsoft really wants us to redistribute their MSI redistributable... I think it's easier for them to keep up to date and patch security vulnerabilities.

Thank you! It seems reasonable for various reasons to not have every application ship their own copy of the dlls. I'm going to check out `burn` as you mentioned or maybe just tell people to install the vcruntime on their own.

nbolton:
> just tell people to install the vcruntime on their own

That's what we're doing with Deskflow since there's no `burn` support for CPack-WiX right now.

For Synergy we configure full-blown WiX manually like you're doing, and we use a naughty merge module, but we plan to move to `burn` soon and bundle the MSVC redist MSI and be good little Windows developers.

PedroHLC:
> Crazy how difficult this stuff is on windows compared to Linux ...

I recommend taking a look at Tauri's codebase, they have .MSI packaging working (with extra customization steps that you could discard), linking rust code properly, and working through cross compilation.

Issue #298, opened 2025-06-11
Wrong Path In lan-mouse.service File


huangsijun17:
The installation path in the README is `/usr/local/bin/lan-mouse', like:

https://github.com/feschber/lan-mouse/blob/b8063a813884b7444fe567deee8f61ff7931d351/README.md?plain=1#L99

The path in `lan-mouse.service` is `/usr/bin/lan-mouse`, which makes the service unable to start:

https://github.com/feschber/lan-mouse/blob/b8063a813884b7444fe567deee8f61ff7931d351/service/lan-mouse.service#L9

PR #299, opened 2025-06-11
Fix The Wrong Path In systemd Config File

huangsijun17:
Modify to the address in README.

fix: https://github.com/feschber/lan-mouse/issues/298#issue-3135301493


feschber:
The service file is used for example in the AUR and other packages. This would break all of those.
If you really want to manually install, you will have to adapt the service file.

I would accept a hint in the README about this, but we can not change the service file.

PR #308, opened 2025-07-04
Add instructions to README for installing on Fedora

ayykamp:
I recently [added lan-mouse to the Terra repo](https://github.com/terrapkg/packages/pull/5659) which can be used in Fedora.

Thought I would add a note to the README to make it easier for other Fedora users.

Edit: I also fixed a broken link.


feschber:
Thank you and sorry for the late reply! Lgtm

PR #327, opened 2025-10-08
implement (small, up to 4k) clipboard sharing over DTLS

dnakov:
Adds bidirectional clipboard sharing between all connected machines.
Clipboard sharing is enabled by default. To disable, add to `~/.config/lan-mouse/config.toml`:
`enable_clipboard = false`

4KB max, no chunking, text only

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.1"
+version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
 name = "aead"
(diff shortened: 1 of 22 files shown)


feschber:
A few things - we dont have to fix any of this right away:

- I'm not a fan of polling the clipboard. This seems a bit wasteful and it seems like there should be a better way? How do clipboard monitoring tools do it?
Maybe, it would make sense to get the clipboard contents upon entering a client device? This would of course not synchronize the clipboard when copied though.

- It might be worth having ClipboardEvent be a direct member of the `input_capture::CaptureEvent` enum instead of `input_event::Event` and split `input_event::Event` into `input_event::{InputEvent,ClipboardUpdate}` or something. This way we could drop the dummy implementations on all of the `input_emulation` backends that do not support clipboard sharing.

- I would prefer to keep `clipboard_monitor` and `clipboard_emulation` out of the service struct and instead move them into the `lan_mouse::emulation::Emulation` / `lan_mouse::capture::Capture` structs respectively.
We should not need a ClipboardReceived Event on the emulation side and could instead handle it like any other emulation event.
ClipboardReceivedEvents coming from the input capture / connection via `ICaptureEvent` could then be send to the emulation via an additional function `lan_mouse::emulation::Emulation::update_clipboard()` and `EmulationRequest::UpdateClipboard(ClipBoardEvent)` in `lan_mouse::emulation` instead of being handled directly in the service.

Let me know your thoughts!

funkyfuture:
imo the feature should be off by default.

feschber:
> imo the feature should be off by default.

definitely going to be configurable via the UI

Issue #156, opened 2024-07-05
Suggestion: Pointer speed/acceleration


asahilina:
Using lan-mouse between machines running KDE Plasma, the mouse feels significantly slower/more sluggish on remote machines than on the local machine. Messing with the mouse settings, I think the issue is that there is no pointer acceleration with remote machines (turning it off locally in the System Settings gives similar behavior). But there is also somewhat of a difference in perceived linear speed, probably due to having different screen scale factors (and physical screen size).

Am I guessing right that the desktop environment acceleration behavior is not available to lan-mouse due to the way it works? If so, it would be nice if lan-mouse could implement acceleration internally, and perhaps also a speed control. Maybe this could be configured in the config file, per-machine-connection to allow different acceleration/speed profiles depending on what machine is the server and what machine is the client? (I have a weird setup where two different host machines can be the input controllers for a single shared mouse-less client machine, so it would not be out of the question to want different speed factors for the shared machine depending on which machine is the active host, since they have different OSes and different mice and might have different behavior)

feschber:
What I definitely want to avoid is amplification of acceleration across devices (pointer acceleration is captured and then accelerated again when being emulated remotely).

In my opinion the only correct thing to do is to send unaccelerated motion to the compositor and let the compositor handle acceleration instead of `lan-mouse`, since the compositor is the only component that can correctly identify if accelerated or unaccelerated motion is required (through interfaces like [relative-pointer-unsatble-v1](https://wayland.app/protocols/relative-pointer-unstable-v1), which report both accelerated and unaccelerated events, letting applications chose what they want).
In other words: Sending accelerated motion _to_ the compositor (and this means any artificial acceleration added by lan-mouse either on the capture or emulation side) would make it impossible for the compositor to infer the original unaccelerated motion.

So ideally, we want to send unaccelerated motion, but also tell the compositor, that it is infact unaccelerated, which I tried to do wherever possible, though in reality most backends do not yet do it correctly.

As for KDE -> KDE, you are (almost certainly) using the `layer-shell` capture backend, which is explicitly using `dx_unaccel` / `dy_unaccel`, so this is working as intended.

[...]

PR #339, opened 2025-10-29
WIP: MacOS input consistency, request for comment

NeoTheFox:
Hello! 
I noticed a huge input inconsistency when going from Sway -> MacOS:
1) The inverted scrolling (despite natural scrolling being disabled in MacOS) 
2) Mouse sensitivity being extremely low (despite both systems using the exact same resolution)

I'm not a Rust developer, so I don't know what would be the best way to approach it - these are just tiny changes that made my systems feel consistent. 
Ideally, the issue is somehow related to DPI and reading the system config, but I propose adding two new config options:
1) mouse_mod - a float value that the actual relative movement gets multiplied by on a given client
2) invert_scroll - a simple flag that inverts the scrolling direction so that you can make it consistent across platforms.

Right now the input-emulation is split away from the config, so I couldn't figure out a way to add these options without excessive coupling. Any comments are welcome.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,10 +1,15 @@
 # example configuration
 
 # configure release bind
-release_bind = [ "KeyA", "KeyS", "KeyD", "KeyF" ]
+release_bind = ["KeyA", "KeyS", "KeyD", "KeyF"]
 
 # optional port (defaults to 4242)
 port = 4242
+
+# linear mouse acceleration
+mouse_mod = 1.0
+# invert scrolling direction
+invert_scroll = false
 
 # list of authorized tls certificate fingerprints that
 # are accepted for incoming traffic
--- a/input-emulation/Cargo.toml
+++ b/input-emulation/Cargo.toml
@@ -9,19 +9,21 @@ repository = "https://github.com/feschber/lan-mouse"
 [dependencies]
 async-trait = "0.1.80"
 futures = "0.3.28"
+float-derive = "0.1.0"
 log = "0.4.22"
 input-event = { path = "../input-event", version = "0.3.0" }
 thiserror = "2.0.0"
+serde = { version = "1.0", features = ["derive"] }
 tokio = { version = "1.32.0", features = [
-    "io-util",
-    "io-std",
-    "macros",
-    "net",
-    "process",
-    "rt",
-    "sync",
-    "signal",
-    "time"
+  "io-util",
+  "io-std",
+  "macros",
+  "net",
+  "process",
+  "rt",
+  "sync",
+  "signal",
+  "time",
 ] }
 once_cell = "1.19.0"
 
@@ -29,19 +31,19 @@ once_cell = "1.19.0"
 bitflags = "2.6.0"
 wayland-client = { version = "0.31.1", optional = true }
 wayland-protocols = { version = "0.32.1", features = [
-    "client",
-    "staging",
-    "unstable",
+  "client",
+  "staging",
+  "unstable",
 ], optional = true }
 wayland-protocols-wlr = { version = "0.3.1", features = [
-    "client",
+  "client",
 ], optional = true }
 wayland-protocols-misc = { version = "0.3.1", features = [
-    "client",
+  "client",
 ], optional = true }
 x11 = { version = "2.21.0", features = ["xlib", "xtest"], optional = true }
 ashpd = { version = "0.11.0", default-features = false, features = [
-    "tokio",
+  "tokio",
 ], optional = true }
 reis = { version = "0.5.0", features = ["tokio"], optional = true }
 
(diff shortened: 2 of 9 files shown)


feschber:
Ideally, I'd want to do this as a general pre / post-processing step (#156).
This means, I'd like it to be applied directly in the [`consume()` function](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/input-emulation/src/lib.rs#L139) before being passed to the device specific backends rather than any platform specific code.

The configuration options would need to be passed to the emulation either when instantiated [src/service.rs#L118](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/src/service.rs#L118) or preferably as a new [`FrontendRequest` variant](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/lan-mouse-ipc/src/lib.rs#L223) which could be used to modify these values in the [`InputEmulation` struct](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/input-emulation/src/lib.rs#L72) at runtime.

Frontend requests are handled [here](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/src/service.rs#L179). We would need a new function on the [emulation](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/src/emulation.rs#L64) to pass the request to the actual input-emulation task in [`EmulationProxy`](https://github.com/feschber/lan-mouse/blob/f91b6bd3c1576d57e743611ff426765f6b02e4c4/src/emulation.rs#L218).

Finally, the frontends (gtk and cli) could then be adjusted to generate these Requests.

Let me know if you feel comfortabl implementing this (or any part of it would be fine, really). There is quite a few components involved unfortunately.

I'm considering moving the whole input-emulation proxy that handles disabled input-emulation state into the input-emulation create eventually, which might make things like this easier to implement.

NeoTheFox:
Thank you for the input! 
I'll look into implementing it this way then

NeoTheFox:
I've partially implemented the mouse_mod and created a new structure for InputConfig, but since now the input is consistent due to #346, is the scrolling inversion the only logical input config to have? Or would querying the natural scrolling config option for the MacOS itself be a better option?

feschber:
Lets look at it this way: I would consider the inconsistent movement and inverted scrolling bugs. However, regardless of this being fixed, it would still be nice to have configurable multiplier / touchpad inversion (#156).

My general idea behind lan-mouse is that everything should feel as though the input devices were directly attached to the target. Unfortunatelly the reality is that in many cases this is not how it currently works. (E.g. MacOS and Windows capture accelerated mouse input right now).

For the touchpad this would mean that the scrolling direction should be the same as on the physical touchpad of the target device - reading the natural scrolling setting somehow may very well be the correct way to do this, yes.

So really, we have two things to do here in my eyes:
1. Make the touchpad behave correctly on macos (as described above)
2. Implement input postprocessing as in configurable multiplier / inverted touchpad, etc. (maybe at some point there could be configurable keymappings as well)

Now which of these you want to implement is totally up to you. If you say, you are happy with the state it is in right now or simply want to invert the default scrolling direction, that'd be fine as well.

NeoTheFox:
My initial idea was implementing a modifier for the mouse (one can call that linear acceleration), but if we're  going to call this acceleration then a it's a more [complex](https://wayland.freedesktop.org/libinput/doc/latest/pointer-acceleration.html) topic, and different devices expect different profiles. But since I'm already working on this we can call it a "hack" for now and I'll just finish doing what I'm doing. 

The natural scrolling thing comes down to being opinionated, so the inversion flag is still a good idea. Since MacOS (just like some other compositors mentioned in #156) doesn't allow setting any pointer configuration to the virtual input itself one can argue that someone might prefer having the natural scrolling on the touchpad, when used on its own, but having it off when using lan-mouse to stay consistent with other computers. 

Rebasing.

feschber:
Ideally, every capture backend would capture unaccelerated input and every emulation backend would emulate the input as though it was coming from a phyiscal device. Then all the input settings from the target device should carry over and things would work as expected. Since this is not the case and may never be in some instances, I guess it is a good idea to make these things configurable.

NeoTheFox:
Implemented cleanly in a different branch, closing this

feschber:
Why not push to this branch?

NeoTheFox:
Dirty history due to my first hacky implementation

feschber:
It's no problem but for the future: You can simply work on a different local branch, then hard reset your other branch to the new one and force push the changes. Or you could setup the new branch to track the remote one and then force push.

NeoTheFox:
Ah, sorry, I've done the first thing that came to mind really

Issue #343, opened 2025-10-31
Odd mismatching vertical / horizontal sensitivity


FabulousCupcake:
As I was trying multiple software kvm switch applications I noticed the input emulation on windows seems to be more sensitive vertically than horizontally with lanmouse.

Unfortunately I don't have an objective way to measure this.  
The best test I could devise is to repeatedly flick my cursor 45° diagonal between two points in my input-capture screen, and do the same in the input-emulated screen. With lanmouse, I always overshoot vertically (upwards to top-right, downwards to bottom-right).

I realize this issue is probably not too helpful, but I hope it may be useful to track if more people encounter similar problem.

### Setup
Input Capture: Mac, native resolution 3840x2160, logical resolution 3008x1692
Input Emulation: Windows, 2560x1440
Version: 35773dfd07c97fc602ba87d65d1c5fcf7740ebb9

feschber:
I'm pretty sure, this is not the input-emulation on the windows side but rather the capture on MacOS. I'm aware of this problem and currently trying to fix some issues on the MacOS side.

feschber:
Might be related:
https://github.com/feschber/lan-mouse/pull/131#discussion_r1602176713

feschber:
Closing for now. Feel free to reopen, if this doesn't resolve things for you!

FabulousCupcake:
Yes! It is resolved, thank you!

PR #346, opened 2025-10-31
fix inconsistent mouse capture on macos

feschber:
#343 this should improve things.
FYI @NeoTheFox (#339)

### DIFF
--- a/input-capture/src/macos.rs
+++ b/input-capture/src/macos.rs
@@ -1,31 +1,39 @@
 use super::{error::MacosCaptureCreationError, Capture, CaptureError, CaptureEvent, Position};
 use async_trait::async_trait;
 use bitflags::bitflags;
-use core_foundation::base::{kCFAllocatorDefault, CFRelease};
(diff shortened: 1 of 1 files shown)


NeoTheFox:
Yes, that takes care of the mouse movement, can confirm it's consistent now

feschber:
@NeoTheFox can you tell if it is slower than it should be?

NeoTheFox:
It feels slower, so I disabled mouse acceleration in sway (accel_profile "flat"), and now it feels 1:1. So it is consistent with the raw input, but now it has more to do with #156

feschber:
> It feels slower, so I disabled mouse acceleration in sway (accel_profile "flat"), and now it feels 1:1. So it is consistent with the raw input, but now it has more to do with #156

On sway, the unaccelerated input is captured. So if this feels 1:1 to you that should be correct.

At some point, I think I may somehow want to implemen a roundtrip test, as in emulate input and capture it to see if it matches.

PR #353, opened 2025-11-15
feat: add binary cache to instructions

skifli:
Adds instructions to add the binary cache to the trusted substituters. I wouldn't have known it existed if I hadn't specifically searched for it in the pull requests tab where it was mentioned in no. 82.


feschber:
Sorry about the late response. Yeah, it seems it isn't mentioned anywhere.

PR #371, opened 2026-02-04
fix: remote key-up on triggered release

### LINKED ISSUE #369: release_bind not clearing modifiers?
I have 2 linux machines running niri (it's niri mainline where they just added libei support) and dank material shell (using dms lock screen).  When I lock a client lan-mouse machine the input gets stuck on that remote machine (lock screen grabbing all input I suspect) so I've configured a release_bind to `Meta+z` to release the remote focus back to my other computer.

The issue is that after doing this, number keys don't work on the remote machine.  It seems like the Meta key up event isn't getting sent on lan-mouse release?  The number keys on the remote computer act like meta is still down, pressing `1` behaves like `Meta+1` and switches niri workspaces.  The end result is that I'm trying to unlock from the lock screen and can't type my password.

```
release_bind = [ "KeyLeftMeta", "KeyZ" ]
```

I'm tinkering around with the code now, mainly just copilot trying out some things in the release function to send key-ups before ungrabbing, but the first couple quick takes weren't helpful.

jonstelly:
Fixes #369.  Issue was that on using a hotkey to trigger remote release (need to do this on a wayland lock screen to return cursor/focus to my main computer), I bound release to Meta+z and the Meta key would be left down/stuck on the remote machine.

Steps to repro:
1. set release_bind in config using some modifier (meta+z is what I used)
2. attempt to release focus from a remote computer with this binding
3. modifier key ends up stuck/down

NOTE: I'm an old developer but pretty new to rust.  This is a pretty small PR but if there's anything you'd like changed, just let me know

### DIFF
--- a/src/capture.rs
+++ b/src/capture.rs
@@ -362,7 +362,13 @@ impl CaptureTask {
     }
 
     async fn release_capture(&mut self, capture: &mut InputCapture) -> Result<(), CaptureError> {
-        self.active_client.take();
+        // If we have an active client, notify them we're leaving
+        if let Some(handle) = self.active_client.take() {
+            log::info!("sending Leave event to client {handle}");
+            if let Err(e) = self.conn.send(ProtoEvent::Leave(0), handle).await {
+                log::warn!("failed to send Leave to client {handle}: {e}");
+            }
+        }
         capture.release().await
     }
 }


simonm:
Hyprland (current) -> macOS (Sequoia). I was constantly having problems when changing focus between machines. This PR fixed the remaining issues I was having with lan-mouse. Now it seems rock-solid and stable. Thank you!

feschber:
LGTM! Thanks.

I missed this when I rewrote the client server architecture with encryption.

Issue #74, opened 2024-01-15
 KDE prompts Remote Control Request muliple times and answer doesn't matter.


CupricReki:
I'm getting multiple Remote Control Request every time I launch lan-mouse from the CLI. 

## Log 
```
❯ lan-mouse
[2024-01-15T14:52:24Z INFO  lan_mouse::config] using config: "/home/cupric/.config/lan-mouse/config.toml"

(lan-mouse:1668930): Gtk-WARNING **: 09:52:24.175: Unknown key gtk-modules in /home/cupric/.config/gtk-4.0/settings.ini
[2024-01-15T14:52:24Z INFO  lan_mouse::config] using config: "/home/cupric/.config/lan-mouse/config.toml"
[2024-01-15T14:52:24Z INFO  lan_mouse] Press Ctrl+Alt+Shift+Super to release the mouse
[2024-01-15T14:52:24Z INFO  lan_mouse::consumer] wayland backend not available: the requested global was not found in the registry
[2024-01-15T14:52:24Z INFO  lan_mouse::producer] libei event producer not available: not implemented
[2024-01-15T14:52:24Z INFO  lan_mouse::producer] using layer-shell event producer

(lan-mouse:1668930): Adwaita-WARNING **: 09:52:24.261: Using GtkSettings:gtk-application-prefer-dark-theme with libadwaita is unsupported. Please use AdwStyleManager:color-scheme instead.
```
![KDE Screenshare](https://github.com/feschber/lan-mouse/assets/6509200/882e06df-058b-40be-895f-31baa2a798ae)
```
[2024-01-15T14:52:34Z INFO  lan_mouse::consumer] libei not available: Portal request failed: org.freedesktop.zbus.Error: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such method 'ConnectToEIS' in interface 'org.freedesktop.impl.portal.RemoteDesktop' at object path '/org/freedesktop/portal/desktop' (signature 'osa{sv}')
```
Remote Control Request prompt shows a second time.
```
[2024-01-15T14:52:40Z INFO  lan_mouse::consumer] using xdg-remote-desktop-portal event consumer
[2024-01-15T14:52:49Z INFO  lan_mouse] terminating service
[2024-01-15T14:52:49Z INFO  lan_mouse::server] terminating service
[2024-01-15T14:52:49Z INFO  lan_mouse::server] terminating gracefully...
```
## Notes
[...]

PR #383, opened 2026-02-10
Handle the RemoteDesktop portal restore token correctly

whot:
For a session to actually persist, we need to request a persistence mode which we already do. The portal then returns a restore-token (in the form of an uuid) to us as part of the response to Start.

This token must then be passed into the *next* session during SelectDevices to restore the previous session.

The token is officially a single-use token, so we need to overwrite it every time. In practise the current XDP implementation may re-use the token but we cannot rely on that.

Reading and writing the token is not async since we expect them to be uuid-length.

### DIFF
--- a/input-emulation/src/libei.rs
+++ b/input-emulation/src/libei.rs
@@ -1,7 +1,8 @@
 use futures::{StreamExt, future};
 use std::{
-    io,
+    env, fs, io,
     os::{fd::OwnedFd, unix::net::UnixStream},
+    path::PathBuf,
     sync::{
         Arc, Mutex, RwLock,
         atomic::{AtomicBool, Ordering},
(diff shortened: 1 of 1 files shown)


whot:
Formatting errors fixed, should pass the tests now.

feschber:
Thank you so much for this! Should fix #74.

On a related note: Does this exist for the input-capture portal as well at this point?

feschber:
Working as intended!
I will fix this in the [remote-desktop backend](https://github.com/feschber/lan-mouse/blob/main/input-emulation/src/xdg_desktop_portal.rs) as well!

nbolton:
> On a related note: Does this exist for the input-capture portal as well at this point?

He's working on it. Here's his draft PR for Deskflow: https://github.com/deskflow/deskflow/pull/9415

FYI, he's also got a Wayland server to non-Wayland client copy/paste PR: https://github.com/deskflow/deskflow/pull/9431

whot:
> I will fix this in the [remote-desktop backend](https://github.com/feschber/lan-mouse/blob/main/input-emulation/src/xdg_desktop_portal.rs) as well!

Ooops, i completely missed that one, sorry about that. Thanks for fixing it though..

As @nbolton already said: it's in the works but currently spread across multiple repos which all need to be built from the respective devel branches. https://github.com/flatpak/xdg-desktop-portal/pull/1898 has the list of all repos in the top comment.

nbolton:
> [flatpak/xdg-desktop-portal#1898](https://github.com/flatpak/xdg-desktop-portal/pull/1898) has the list of all repos in the top comment.

I think I ended up building like 5 libs on GNOME :smile: (still need to finish that task off)

Issue #357, opened 2025-12-12
[Linux client + MacOS server] Sticky modifier keys


kiwiz:
Client (IO source): KDE/Wayland, Ubuntu, commit: 3922b45bd92f07f62e8af84bb9471abf8b0873a3
Server: MacOS, commit: 640fa995a4ae3c5a0aaaafe04857930979edc220

When performing key combinations (ex: meta+tab), the modifier key appears to get stuck down. Hitting the key again resets the state. FWIW, MacOS -> MacOS works perfectly fine.

kiwiz:
Mac -> Mac
```diff
+ [2026-01-12T23:35:37Z WARN  input_emulation::macos] Event::Keyboard Keyboard(Key { time: 0, key: 125, state: 1 })
+ [2026-01-12T23:35:37Z WARN  input_emulation::macos] update_modifiers Cell { value: XMods(0x0) } 125
+ [2026-01-12T23:35:37Z WARN  input_emulation::macos] key event: 55 1
- [2026-01-12T23:35:37Z WARN  input_emulation::macos] modifiers updated: XMods(Mod4Mask)
...

+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] Event::Keyboard Keyboard(Key { time: 0, key: 15, state: 1 })
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] update_modifiers Cell { value: XMods(Mod4Mask) } 15
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] update_modifiers Cell { value: XMods(Mod4Mask) } 55
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] key event: 55 0
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] key event: 48 1

+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] Event::Keyboard Keyboard(Key { time: 0, key: 15, state: 0 })
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] update_modifiers Cell { value: XMods(Mod4Mask) } 15
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] update_modifiers Cell { value: XMods(Mod4Mask) } 48
+ [2026-01-12T23:35:39Z WARN  input_emulation::macos] key event: 48 0

[...]

PR #385, opened 2026-02-11
Linux -> macOS fixes

kenichi:
* address #357 - When handling a KeyboardEvent::Key for a modifier key, emit a CGEventType::FlagsChanged event so macOS updates its system modifier state, rather than relying solely on a KeyboardEvent::Modifiers event that Linux/libei never sends.

### DIFF
--- a/input-emulation/src/macos.rs
+++ b/input-emulation/src/macos.rs
@@ -418,10 +418,7 @@ impl Emulation for MacOSEmulation {
                     };
                     let is_modifier = update_modifiers(&self.modifier_state, key, state);
                     if is_modifier {
-                        modifier_event(
-                            self.event_source.clone(),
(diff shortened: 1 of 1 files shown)


kiwiz:
Confirming that this fixes macOS for me, thx!

IMO, #347 is preferable over hardcoding the reversal here.

kenichi:
@kiwiz glad to hear it!

i can rebase this to just the modifier keys event commit if there's a better fix for scrolling, whatever works best @feschber

feschber:
I don't really have any strong opinions on whether or not we should hard code either scroll direction (when #347 is merged, it ultimately doesn't matter).
However, I'd prefer to have a separate PR for it, if you want it to be included, so this one is only about the modifier event emulation.

PR #392, opened 2026-02-19
fix(macos): forward back/forward mouse buttons in capture and emulation

### LINKED ISSUE #391: Feature Request: MacOS does not handle additional mouse buttons like back and forward
The MacOS implementation has 2 buttons mapped, then an "other" that defaults to middle button. While other systems transmit raw evcodes or have explicitly mapped logic.

tyvsmith:
OtherMouseDown/Up events on macOS carry a button number field that distinguishes middle (2), back (3), and forward (4) buttons. The capture backend was unconditionally mapping all OtherMouse events to BTN_MIDDLE, silently dropping back/forward. The emulation backend had no match arms for BTN_BACK/BTN_FORWARD, causing them to be dropped with a warning.

Fix capture by reading MOUSE_EVENT_BUTTON_NUMBER and mapping 3->BTN_BACK, 4->BTN_FORWARD. Fix emulation by adding match arms for BTN_BACK/BTN_FORWARD and setting MOUSE_EVENT_BUTTON_NUMBER on the emitted CGEvent so macOS apps receive the correct button identity.

Verified on Hyprland Arch host running the latest main build, and macOS Tahoe 26.2 as the client.

Fixes #391

### DIFF
--- a/input-emulation/src/macos.rs
+++ b/input-emulation/src/macos.rs
@@ -421,7 +421,10 @@ impl Emulation for MacOSEmulation {
                             return Ok(());
                         }
                     };
-                    update_modifiers(&self.modifier_state, key, state);
+                    let is_modifier = update_modifiers(&self.modifier_state, key, state);
+                    if is_modifier {
+                        modifier_event(self.event_source.clone(), self.modifier_state.get());
+                    }
                     match state {
                         // pressed
                         1 => self.spawn_repeat_task(code).await,


feschber:
LGTM!

tobiashugo:
This build crashes on Mac as soon as I try to authorize a server. Even with empty input fields crash to desktop.

PR #412, opened 2026-04-04
feat(x11): implement X11 input capture backend

posthum4n-ai:
We (Claude+me) got X11 input working using polling instead of mouse events. For me this works perfectly on X11, is this something we want into main? Tested with Gnome 46 + X11.

I don't have experience with Rust or X11, but I'm super happy with how this works for me.

### DIFF
--- a/input-capture/src/error.rs
+++ b/input-capture/src/error.rs
@@ -137,8 +137,10 @@ pub enum LayerShellCaptureCreationError {
 #[cfg(all(unix, feature = "x11", not(target_os = "macos")))]
 #[derive(Debug, Error)]
 pub enum X11InputCaptureCreationError {
-    #[error("X11 input capture is not yet implemented :(")]
-    NotImplemented,
+    #[error("could not open X11 display")]
+    OpenDisplay,
+    #[error("io error: {0}")]
+    Io(#[from] std::io::Error),
 }
 
 #[cfg(target_os = "macos")]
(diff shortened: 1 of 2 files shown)


feschber:
Is there a good reason not to listen to mouse events rather than polling?
I'm not a huge fan of the "do something every 2ms" sort of thing.

Technically there should also be no reason to even spawn a separate thread (this is something I'm willing to accept and improve upon later)

posthum4n-ai:
I agree because I hate polling too, I was under the impression that doing this event based was much harder (I assumed this because issues with X11 mouse events was the reason it wasn't implement for X11 yet). 

I did notice an issue with log spam yesterday, that needs to be fixed first anyway (for myself at least :)

Separate thread probably is over-engineered, it kinda of made sense to me because it won't affect main thread/mouse responsiveness. 

Because of the position polling method the cursor also needs to have offsets from border when re-entering a display.

Feel free to stash for now, I just wanted to show that it technically works with X11 (after switching back from Wayland), but not sure if it fits your application.

posthum4n-ai:
I'll close it for now, okay.

feschber:
I'm open to merging this and improving it later.
Its certainly much better than no backend at all. I just wanted to know your reasoning first.

posthum4n-ai:
It's cool, I noticed another issue where my mouse got stuck, but I think that happened because of some weird glitch that duplicated my displays... Hopefully I can fix mainly the logging spam before I add or update this PR!

PR #415, opened 2026-04-23
fix: negate macOS scroll delta to match wl_pointer/libei sign

jondkinney:
## Summary

macOS CGEvent scroll deltas use the opposite sign convention from `wl_pointer.axis` (and the libei capture backend used on Linux hosts). On macOS, a positive delta means "content moves up" (finger-up gesture under natural scrolling), while on Wayland a positive axis value means "scroll-down direction". The wire protocol value was therefore landing on the sink with the sign flipped relative to what other capture backends produce, so scrolling felt backwards whenever the source machine was macOS.

This normalizes the sign at capture time in ``input-capture/src/macos.rs`` so the wire protocol stays source-agnostic — both continuous (trackpad/Magic Mouse point-delta) and line-based (mouse-wheel) scroll values are negated to match wl_pointer/libei semantics.

### Before / after
- **Before**: scrolling from a macOS source with another lan-mouse client on Linux (wlroots emulation backend) scrolled the opposite direction from what the user gestured on the Mac.
- **After**: gesture direction on macOS matches content direction on the Linux sink, identical to behavior with a libei/Linux source.

## Test plan
- [x] Tested locally with a Mac (Dygma Defy + Magic Trackpad) as the source and an Arch + Hyprland host (wlroots sink) as the target. Scroll direction now matches gesture direction on both machines.
- [ ] Would appreciate confirmation from another macOS user that direction feels correct with both natural-scroll on and off in System Settings.
- [ ] Line-based (non-continuous) scroll path is negated symmetrically; no test environment with a PS/2-style wheel mouse on Mac to verify, but the change is trivially the same sign flip applied to the ``V120_STEPS_PER_LINE * v`` calculation.

## Notes
The fix deliberately lives in the macOS capture path rather than on the emission side, so the wire format stays consistent with what the libei capture backend already produces. No sink-side change is needed and Linux→Linux scrolling is unaffected.

### DIFF
--- a/input-capture/src/macos.rs
+++ b/input-capture/src/macos.rs
@@ -357,6 +357,10 @@ fn get_events(
             })))
         }
         CGEventType::ScrollWheel => {
+            // macOS CGEvent scroll deltas use the opposite sign convention
+            // from wl_pointer/libei (positive = scroll-down direction).
+            // Negate so the forwarded values match the wire protocol
+            // convention used by the libei/Linux capture backend.
             if ev.get_integer_value_field(EventField::SCROLL_WHEEL_EVENT_IS_CONTINUOUS) != 0 {
                 let v =
                     ev.get_integer_value_field(EventField::SCROLL_WHEEL_EVENT_POINT_DELTA_AXIS_1);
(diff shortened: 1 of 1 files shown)


jondkinney:
Closing — the reported bug was misdiagnosed on my end.

Follow-up testing showed that macOS applies the "Natural scrolling" System Settings toggle (`com.apple.swipescrolldirection`) **before** the CGEvent reaches the `CGEventTap` callback. So the raw `SCROLL_WHEEL_EVENT_POINT_DELTA_AXIS_*` / `SCROLL_WHEEL_EVENT_DELTA_AXIS_*` values are preference-dependent, not a fixed physical convention.

Verified on macOS 15.5 Sequoia:

- Mac source (natural scrolling ON) → Linux sink (wlroots/libei emulation): scrolls correctly **without** any sign flip. This is what the no-change mainline does today.
- Mac source (natural scrolling OFF) → same Linux sink: scrolls backwards, which is what I originally reported.

So this PR's unconditional negation fixes the natural-scroll-OFF case but would break the default (natural-scroll-ON) case for every other user. Wrong fix.

Sorry for the noise.

PR #416, opened 2026-04-23
macOS: fix missing symbolic icons and improve app icon

jondkinney:
## Summary

Two macOS-specific fixes:

- **Symbolic icons render as "image-missing" placeholders.** The GTK frontend references Adwaita symbolic icon names (`edit-copy-symbolic`, `auth-fingerprint-symbolic`, `network-wired-symbolic`, `dialog-warning-symbolic`, etc.) but the Adwaita icon theme isn't installed on macOS by default. Bundling `adwaita-icon-theme` via Homebrew doesn't reliably fix it either — an `.app` launched from Finder inherits a minimal `XDG_DATA_DIRS` that doesn't include `/opt/homebrew/share`, so GTK can't find the theme at runtime.

  Fix: bundle the needed symbolic SVGs directly into the embedded gresource under the standard `scalable/{actions,devices,places,status}/` hicolor layout. The existing `IconTheme::add_resource_path("/de/feschber/LanMouse/icons")` call already tells GTK to search this prefix, so no Rust changes are needed. This makes the app self-contained on macOS with no runtime dependency on a system icon theme.

- **Dock/Finder app icon looks oversized and square.** The existing `makeicns.sh` produced a 1024×1024 icon with no squircle background and no transparent padding, so macOS rendered it noticeably larger than first-party apps and without the rounded-square shape users expect.

  Fix: rewrite the script to follow Apple's Big Sur+ icon template — 1024 canvas, 824 white squircle centered with 100px transparent padding, artwork rendered at 560 centered inside the squircle, ~22.5% corner radius. Use `rsvg-convert` to rasterize the SVG reliably (ImageMagick crops this particular SVG) then composite onto the squircle.

After changes - in the Dock:

CleanShot 2026-04-23 at 14 04 58@2x

CleanShot 2026-04-23 at 15 35 15@2x

## Test plan

- [x] Build the GTK frontend on macOS — symbolic icons now render correctly for hostname-copy, certificate fingerprint, authorize, emulation/capture status rows, connection rows (network-wired, edit-delete), etc.
- [x] `scripts/makeicns.sh` produces a valid `target/icon.icns`; bundled app icon appears in the Dock/Finder with proper squircle shape and sizing consistent with other macOS apps.
- [ ] Linux/Windows builds unchanged — `makeicns.sh` is only invoked from the two macOS jobs in `.github/workflows/release.yml`, and the gresource addition is pure data (new files + XML entries, no code changes).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #417, opened 2026-04-24
Macos menubar app

jondkinney:
This includes commits from #416 as well so we can likely close that other PR. 

The app will now properly prompt for accessibility permissions, and monitor for them being granted. Once granted the app will prompt to be restarted for the new permissions to take effect. That auto-relaunch flow and manual launches open the GUI window automatically, but closing that window leaves the app running in the menubar. And if it's set to auto-launch, then the GUI doesn't show on a reboot, etc. 

If you revoke the accessibility permissions, the app quits itself since not quitting would leave the input system in a broken state where the cursor still moves but clicking doesn't work, nor do keypresses. Once revoked, a re-launch will re-prompt for accessibility permissions.

# Menubar 
CleanShot 2026-04-24 at 10 58 46@2x

# Dock
CleanShot 2026-04-23 at 14 04 58@2x

# Granting permissions

CleanShot 2026-04-24 at 11 03 43@2x

## Restart required

CleanShot 2026-04-24 at 11 05 22@2x

## Restarted

CleanShot 2026-04-24 at 11 06 44@2x

## Download this PR's build for macOS for testing

[Lan Mouse.zip](https://github.com/user-attachments/files/27183767/Lan.Mouse.zip)

Because this is an ad-hoc signed bundle, you will need to remove quarantine before it will open

```
xattr -rd com.apple.quarantine "/Applications/Lan Mouse.app"
```


jondkinney:
@feschber I think this is ready for a look if you're game! I don't plan to make any more changes unless something else breaks. It's been working really well in my testing the last two days.

feschber:
Thank you, I will take a look! Been very busy the past couple of days...

feschber:
I don't have any major gripes! Looked over everything. I would say we can merge this once you downgrade libadwaita back to 1.1

jondkinney:
@feschber thanks for the feedback! I addressed it in the latest commit. The last thing that I've found in my testing that I'd love to refine, but am not sure if would cause regressions or have unintended consequences, is to have some sort of "auto-release" of the mouse so that moving between lock screens on macOS and Arch doesn't cause the cursor to get "stuck". It's easy to release it with the `⌘ + ⌥ + ⌃ + ⇧` chord, but if someone doesn't know about that, they'd feel like it broke. 

The issue is when either system is locked, moving between them can be problematic. Any thoughts? Worth investigating? Even if so, maybe that'd be better for a different PR...

feschber:
Yeah that would be something for a new PR. Its also very dependent on the backend (not sure if possible on layer-shell / arch at all for example).

I'm going to not squash this PR however, since you have a very clean history and quite a lot of changes.

There is also a few issues I think your changes solve, which I'm going to look up

PR #418, opened 2026-04-30
[OLD all-in-one version] - Wall-press auto-release, scroll preferences, GUI singleton, and macOS QoL

jondkinney:
## Summary

Ten independent feature areas, each useful on its own:

1. **Host-lock cross-suppression** — when the host's screen is locked, the cursor refuses to cross to the peer. Standardizes behavior across all platforms (Wayland already enforced this; macOS and Windows didn't) so a half-broken state where the mouse moves to the peer but the keyboard goes to the host's lock screen can no longer be reached. Wayland gets it free from the compositor; macOS polls `CGSessionCopyCurrentDictionary` at cross-decision time; Windows registers for `WM_WTSSESSION_CHANGE`. The complementary direction — *cursor already on the peer* when the host locks — is covered by §2's wall-press auto-release: host-side input hooks keep running under the lock screen, so pushing against the host-adjacent edge of the peer ejects the cursor back to the host without any extra plumbing.
2. **Wall-press auto-release** — capture self-releases when the cursor is held against the host-adjacent edge of the guest, gated by a peer-Leave deadline so it stays a fallback rather than racing the layer-shell handover. New `Bounds` proto event so the host knows the guest's extent and can clamp / warp accurately. Doubles as the user-facing escape hatch for §1's locked-host-while-on-peer case.
3. **Per-pair scroll & motion-sensitivity controls** — each authorized peer's row in Incoming Connections expands to a Natural Scrolling switch and a Mouse Sensitivity spin (range 0.1–5.0). Settings stored on the fingerprint-keyed peer entry; receive-side transform is centralized in `InputEmulation::consume`. New `ProtoEvent::ReceiverSensitivity` so the host's wall-press auto-release model scales by the receiver's multiplier (sub-1.0 sensitivity no longer triggers premature cross-back). Plus the foundational scroll-correctness fixes (macOS classic-direction wire convention, macOS v120 line→tick mapping, wlroots `axis_source` for continuous scroll). Plus mDNS-resolved hostname / IP in each row's subtitle and a copy-able fingerprint row in the expansion. Adapted from #347 by @NeoTheFox / Raidon Chrome (per-pair extension + GUI surfacing + wall-press protocol fix); see §3 for the full comparison.
4. **Cross-platform GUI singleton** — `lan-mouse` launched twice no longer opens a second window. Decoupled from the daemon socket; headless `lan-mouse daemon` workflow unchanged.
5. **macOS QoL** — Cmd+W collapses to menu bar, default height fits modern displays, Auto-Release group label scoped to "outgoing", quit unfreezable (3s daemon SIGKILL + 5s process force-exit backstop), display wakes on incoming input.
6. **Cross-machine cursor position sync** — when control transfers, the destination cursor lands at the visually-corresponding point. Survives mismatched resolutions, fractional/software scaling, fast back-and-forth handovers, and the bootstrap problem on the very first crossing.
7. **Peer version exchange** — `Hello` proto event lets each peer surface the other's build commit hash in the GUI, with a soft-warn color indicator on mismatch. Listen-side mirror so version visibility doesn't depend on outbound reachability. Plus a connect-side `Hello`-echo broadcast and `get_client(active_addr.ip())` fallback that close a race in mDNS-primary scenarios where the peer's actual operational IP differed from DNS-resolved values, leaving the version display silently stuck at "unknown."
8. **Hostname resolution via the OS resolver** — `JKMBP-M4-Max.local` and similar Bonjour names now Just Work because `getaddrinfo` walks `/etc/nsswitch.conf` / mDNS / `/etc/hosts` instead of pure-DNS-only.
9. **Multi-homed DTLS listener** — one listener per local IPv4 instead of a single `0.0.0.0` bind, so reply packets always source from the IP the peer dialed. `if-watch` supervisor with periodic reconciliation handles interface plug/unplug and admin-disable.
10. **mDNS-SD service-order discovery** — each instance advertises a `_lan-mouse._udp.local.` Bonjour service whose TXT record names the OS-preferred-interface IP. Dialer biases toward that IP via a 200ms happy-eyeballs head-start, so multi-homed peers connect via the right interface (Mac service order / Linux default route) without a manual `ips=[…]` workaround.

History is pre-cleaned: each commit single-purpose. Diagnostic markers at `log::info!` (`[wp-begin]`, `[release-warp]`, `[bootstrap]`, `[cursor-pos]`, `mdns: …`, `wall-press …`) are intentional — set `LAN_MOUSE_LOG_LEVEL=warn` to silence.

---

## 1. Host-lock cross-suppression

When the host's screen is locked, prevent the cursor from leaving for the peer — the host's lock screen consumes keyboard events before any capture hook sees them, so a mouse-only-on-peer state is broken-by-design. This section standardizes that prevention across all platforms; previously Wayland enforced it (because the compositor revokes input on lock) but macOS and Windows didn't, so users on those hosts could silently end up in the half-broken state.

Two cases worth distinguishing:

- **"Host is unlocked, cursor on host, user moves toward the edge while host locks"** — refuse to commit the cross. This section's primary job. All three platforms now do it.
- **"Cursor is already on the peer when the host locks"** — eject the cursor back to the host. This is where §2's wall-press auto-release earns its keep: the host-side CGEventTap / WH_MOUSE_LL hook keeps running under the lock screen, so the user's motion still accumulates in the host's `virtual_cursor` model, and pushing against the host-adjacent edge of the peer fires the wall-press release as if everything were normal. Verified end-to-end on macOS host ↔ Linux peer: lock the Mac while cursor is on Linux, push left, cursor returns. Same mechanism works on Windows host (and on Wayland host where the compositor handles it natively). Release-bind chord and peer-side `Leave` are the alternative recovery paths.

**Wayland (`free`)** — the compositor revokes input on layer-shell surfaces when the screen locks, so `wl_pointer.Enter` stops firing AND the in-flight `wl_pointer.Leave` arrives, naturally tearing down whichever direction was in play. No code needed.

**macOS (`8cd86ba` / `9d6f427` / `76f3544`)** — initial attempt was `CFNotificationCenterAddObserver` for `com.apple.screenIsLocked`/`Unlocked`, but those callbacks never fired in practice because lan-mouse's main thread runs the GLib event loop, not a CFRunLoop, and the distnoted mach port attaches to the main thread regardless of which thread called `AddObserver`. Replaced with a direct `CGSessionCopyCurrentDictionary["CGSSessionScreenIsLocked"]` poll evaluated at the moment a barrier crossing is about to commit — ~10–50 µs of XPC to WindowServer, paid only on cross attempts (a few per minute), zero per-event overhead during in-flight capture. Mid-capture lock isn't separately handled and doesn't need to be: §2's wall-press is the user-facing recovery path and it works under the lock screen because the CGEventTap keeps running.

**Windows (`8cd86ba`)** — `WTSRegisterSessionNotification` on the existing message-only window; `window_proc` flips a `HOST_LOCKED` thread-local on `WTS_SESSION_LOCK` / `WTS_SESSION_UNLOCK` and gates `check_client_activation`. Convenience extra: also synthesizes `CaptureEvent::AutoRelease` if a capture is already in flight when the lock fires — same hand-off into §2's release plumbing, so the cursor returns to the host the instant the lock screen comes up rather than waiting for the user to wall-press out. Possible because the WTS notification arrives unconditionally; on macOS the equivalent would require polling on every motion event during capture, and §2's wall-press already covers it.

## 2. Wall-press auto-release

User pushes cursor past the host-adjacent edge of the guest and keeps pushing for N more pixels → capture releases. Threshold tunable in GUI; 0 disables.

**The triggering scenario is two locked screens.** When the peer's screen is locked, its lock screen owns the input pipeline before lan-mouse's capture barrier can fire — so the peer can't detect the user crossing back and never sends `Leave` to the host. The host stays in capture mode indefinitely: every keystroke goes to the (probably-locked) guest, and there's no way to unlock the host without first recovering the cursor via the release-bind chord (Ctrl+Shift+Meta+Alt). Wall-press auto-release breaks this deadlock by treating sustained motion past the host-adjacent edge as an explicit "I want out" signal — without protocol cooperation from the peer, which is the part that's broken.

**Same mechanism doubles as §1's escape hatch.** When the host's screen is locked while the cursor is already on the peer, the host's input hook keeps running under the lock screen, so wall-press still accumulates and still fires. Without it, a user whose host locks mid-capture would have to fall back to the release-bind chord or wait for the peer to fire a `Leave`. With it, "lock the Mac with cursor on Linux, push left, cursor returns" Just Works on every platform that has wall-press configured.

The naive virtual-cursor accumulator runs away if the user holds against the wall, so the protocol-based fix adds `ProtoEvent::Bounds(width, height)` (sent by emulation right after Enter), caches it per-position, and clamps `virtual_pos` to the peer's actual extent. Emulation also warps the cursor on Enter so `virtual_pos = 0` lines up with the guest's actual cursor.

`display_bounds` and `warp_cursor` are added as trait methods on `InputEmulation` and implemented for every backend except `xdg_desktop_portal` (no protocol support — falls back to heuristic-only path).

**Peer-Leave deadline gate (`34605a7`):** wall-press used to fire the moment the threshold was crossed, which raced the peer-side layer-shell handover and only "worked correctly" because the network round-trip beat 200px of physical motion. With the gate, wall-press defers AutoRelease for ~150ms after threshold and cancels if a peer `Leave` arrives in that window. Result: in normal operation the layer-shell handover always wins (no spurious wall-press fires); in true fallback scenarios (peer's layer-shell suppressed, or §1's locked-host-stuck-on-peer case) the deadline elapses and wall-press fires as designed.

Bundled forward-compat fix (`3025422`): we now log-and-skip undecodable peer datagrams instead of disconnecting. Necessary because peers running pre-`Bounds` builds can interop with this branch.

## 3. Per-pair scroll & motion-sensitivity (Incoming Connections rows)

**Per-peer post-processing schema.** Each authorized peer's row in the Incoming Connections list is now an `AdwExpanderRow` with two new controls inside: a Natural Scrolling switch and a Mouse Sensitivity spin (range 0.1–5.0, step 0.1, default 1.0). Settings are stored on `IncomingPeerConfig`, keyed by TLS certificate fingerprint — the only stable cross-session identity for an inbound peer. Persisted to `config.toml`. Backwards-compat `Deserialize` accepts legacy bare-string entries and auto-migrates them to defaults; existing configs keep working.

**Centralized transform.** Both the multiplier (on `PointerEvent::Motion`) and the sign-flip (on `Axis` / `AxisDiscrete120`) live in a single match in `InputEmulation::consume`, keyed by `EmulationHandle`. The platform emulation backends (macos, wlroots, libei, x11, windows, xdg_desktop_portal) go back to platform-mechanics-only — no per-backend `set_natural_scroll` shadow state. Match-arm pattern adapted from #347 by Raidon Chrome / NeoTheFox; ported here to per-handle and credited via `Co-Authored-By` on the relevant commit.

**Wall-press fix via wire protocol.** Sub-1.0 sensitivity used to fire AutoRelease before the receiver's cursor reached the wall, because the host's wall-press accumulator ran against raw deltas while the receiver applied the multiplier. New `ProtoEvent::ReceiverSensitivity { mouse_sensitivity: f64 }` (sent by receiver after `Ack` of `Enter`, plus live on every slider change for active peers) lets the sender's wall-press model scale its accumulator to match the receiver's actual cursor advance. Forward-compatible: old peers silently skip the unknown variant.

**Cross-back cycle preservation.** `EmulationTask`'s `post_processing[addr]` cache outlives `ProxyRequest::Remove` (which fires on every `Leave` and on the heartbeat sweep), so per-pair settings follow the SocketAddr instead of the ephemeral handle that's minted fresh on each cross.

**In-place GTK row updates.** `Window::set_authorized_keys` diffs the incoming map against existing rows and mutates `KeyObject` properties in place, rather than `remove_all` + rebuild. Otherwise every `AuthorizedUpdated` (which fires on every per-pair setting change too) would collapse every expanded row in the list. KeyRow listens to property-notify on the bound KeyObject so widget state tracks in-place mutations without ping-ponging back as fresh user requests.

**mDNS-resolved peer identity in the row subtitle.** On DTLS connect, Service does a reverse-lookup against the existing `PrimaryCache` (`hostname → primary_ip`, populated by mDNS browse events) and stores `last_addr` + `last_hostname` on `IncomingPeerConfig`. Persisted across daemon restarts. Subtitle renders as `mac-mini (192.168.1.42)`, `192.168.1.42`, `mac-mini`, or `(not yet connected)` depending on what's known. The fingerprint moves from the title-row subtitle to a copy-able row at the top of the expansion; Revoke Authorization moves to a destructive row at the bottom.

### Scroll-correctness fixes (foundational, still here)

**macOS capture (`dc09e88`):** Wire convention canonicalized to classic mouse-wheel direction regardless of the user's macOS Natural Scrolling preference (read via `CFPreferencesCopyAppValue`). Without a fixed wire convention, the receiver-side toggle has no stable reference. Opposite of `feschber/lan-mouse#415`, which negated unconditionally and traded the bug for the opposite-preference user.

**wlroots emulation (`a26ae8b`):** Continuous-scroll events (trackpad) were silently dropped by Hyprland/Sway/GNOME-Shell because the `axis()` event lacked a companion `axis_source` in the same `wl_pointer.frame`. Spec calls it a "hint"; in practice it's load-bearing for continuous scroll. One extra emit per axis frame fixes it.

**macOS line→tick mapping (`19d36bf`):** Capture used to map one macOS scroll-line to 40 v120 units (1/3 of a wheel tick), so receivers using the discrete count needed 3+ notches before any scroll registered. macOS already amplifies `SCROLL_WHEEL_EVENT_DELTA` by velocity, so one line should map directly to one full v120 tick. Slow-notch and fast-flick behavior is now symmetric across native-Wayland and XWayland clients.

### Relationship to #347

#347 by @NeoTheFox / Raidon Chrome was the first PR to address mouse sensitivity and scroll inversion in lan-mouse, originally discussed in #339. This PR's per-pair scope, GUI surfacing, and wall-press wire-protocol fix make it a superset: setting all peers to the same values reproduces #347's global behavior, and per-pair handles the multi-sender case where senders on different OSes emit deltas at incompatible scales (Mac trackpad vs Windows high-DPI mouse) that no single receiver-side multiplier can satisfy. Receive-side `consume()` algorithm is borrowed from #347 with `Co-Authored-By` attribution. #435 is the focused-review home for this work and includes a fuller comparison.

## 4. Cross-platform GUI singleton (`60041ae`)

Dedicated `lan-mouse-gui.sock` (Unix) / `127.0.0.1:5253` (Windows), separate from the daemon socket. First GUI binds; later launches connect, send a byte, and exit. Primary forwards into the GTK main loop and calls `window.present()`.

Decoupled from the daemon socket on purpose: `lan-mouse daemon` headless workflow is unchanged. Stale-socket recovery if the primary crashed without cleanup. Defense-in-depth via `app.windows().first()` for the in-process activation path. Unit test covers acquire → signal → re-acquire.

## 5. macOS QoL

- **Cmd+W (`afe9456`)** wires to GtkWindow's `window.close` action, which on macOS hides the window and flips the activation policy to `Accessory` — effectively collapses to the menu bar. Linux/Windows unchanged.
- **Default height 700→1400 (`8a86b9d`)** fits every preference group on first paint on ≥1440px-tall displays. Tiling Wayland WMs ignore `default-height`, so no effect there.
- **"Outgoing Auto-Release" rename (`8b0a169`)** scopes the group label to what it actually controls. Description retightened (`66df7e5`) to frame it as a peer-locked-screen fallback rather than a positive-action feature, since the deadline gate (§2) makes that its actual role.
- **Quit unfreezable (`39752ee`)**: 3s `try_wait()` poll on the daemon child with SIGKILL fallback, plus a process-level `std::thread` 5s force-exit backstop scheduled *outside* the GTK main loop (so a wedged loop can't prevent it). Worst-case quit latency 5s; normal completes in <1s.
- **Display wake on input (`56f828f`)**: emulation backend pokes `IOPMAssertionDeclareUserActivity` whenever a forwarded event arrives, so a peer-driven keystroke or click wakes the macOS display from idle-sleep. Without this, lan-mouse would keep pumping events to a blanked screen.
- **Slider scroll passthrough (`3743edd`)**: cursor over the release-threshold slider no longer eats main-window scroll events. Capture-phase handler suppresses `GtkScale`'s own scroll-to-adjust handler (so the slider value stays put) AND forwards the scroll to the ancestor `ScrolledWindow` (so the page scrolls naturally).

## 6. Cross-machine cursor position sync

The host computes its cursor's position as a normalized fraction `(nx, ny) ∈ [0, 1]` against its own bounds and sends `ProtoEvent::CursorPos { pos, nx, ny }` right after Enter. The receiver scales against its own live bounds and pins the on-axis dimension to the entry edge. Self-sufficient — works on the very first crossing, no `Bounds` round-trip needed.

**One architectural change to flag for review:** the release path is now split. `release_capture` (release-bind chord, backend auto-release) computes a host-side cursor warp from the cached `virtual_cursor`. `release_capture_handover` (used when the peer takes over via `ReleaseNotify` or `ProtoEvent::Leave`) skips the host warp so the peer's authoritative `CursorPos` is the only signal seating our shared cursor.

## 7. Peer version exchange (`29da4d8`)

`ProtoEvent::Hello { commit: [u8; 8] }` carries each peer's `shadow_rs` `SHORT_COMMIT` once per session. Sender fires immediately after DTLS auth; listener mirrors the event back so the connect side's `receive_loop` populates `ClientState::peer_commit` for the right handle. Disconnect path clears it.

Each outgoing-connection row's collapsed subtitle renders match status with Pango-colored markup: green when commits match, orange when mismatched or when the peer hasn't sent `Hello` (older build). Soft-warn only — version mismatch never refuses traffic. The local commit reaches the GTK frontend (separate process) via an explicit `local_commit` parameter on `lan_mouse_gtk::run`, stashed in a `OnceLock` so per-row UI can compare against each peer's hash without an IPC round-trip.

`EventType::Hello` is appended to the enum so existing IDs are untouched. Old peers hit the existing `InvalidEventId` skip path from `3025422` and silently ignore the event — backward interop preserved.

**Listen-side mirror (`1ea7148`):** the original implementation read `peer_commit` only off the outgoing-connect path, on the assumption that bidirectional setups always have a working outbound connection in both directions. That assumption broke the moment any direction's outbound was down (e.g. peer's TCP listener temporarily not bound) — version display silently said "unknown" while the peer was happily sending events to us inbound. New `EmulationEvent::PeerHello { addr, commit }` variant fired from the listen-side Hello handler; service maps `addr → ClientHandle` via `client_manager.get_client(addr)` and stamps `peer_commit` exactly like the connect path. Version visibility is now independent of outbound reachability.

**`get_client` matches `active_addr.ip()` too:** the listen-side mirror's `get_client(addr)` lookup matched only against `s.ips` (DNS-resolved IPs ∪ `fix_ips`). With mDNS-primary peers, the actual operational IP often differs from DNS-resolved values — the dialer connects at the mDNS primary, but `s.ips` still has only the stale DNS entry. Mac's incoming `Hello` arrives from the mDNS-primary IP, doesn't match `s.ips`, gets dropped on the floor, and the GUI silently stays at "unknown." Now `get_client` also matches against the client's recorded `active_addr.ip()` (set whenever the outbound dial succeeds), which the dialer always populates with the actually-connected addr.

**Connect-side `Hello` echo broadcasts to the GUI:** historically a fire-and-forget mutation. The `Hello`-echo received over the outbound DTLS connection was written to `client_manager.peer_commit` but never made it to a `FrontendEvent::State` broadcast — closes a race where the listen-side `PeerHello` fired before the outbound dial populated `active_addr` and the listen-side path's `get_client(addr)` returned None, leaving the GUI stuck at "unknown" until next daemon restart. Plumbed via a new `ICaptureEvent::PeerCommitUpdated(handle)` so Service can `broadcast_client` whenever the connect-side stores a fresh commit.

## 8. Hostname resolution via the OS resolver (`9ce3847`)

`hickory_resolver::TokioResolver` only consults `/etc/resolv.conf` and queries upstream DNS servers — which means it can't see `/etc/hosts`, mDNS/Avahi/Bonjour, NetBIOS, or anything else in the system's full name-resolution stack. On a typical home LAN there's no DNS server that knows about peer machine names, so users had to fall back to typing IP addresses, which broke the moment they moved their setup to a different network.

Swap to `tokio::net::lookup_host`, which calls `getaddrinfo`. That walks `/etc/nsswitch.conf` on Linux (picking up Avahi-resolved `.local` names, `/etc/hosts`, and DNS), uses Bonjour for `.local` on macOS, and the full Windows resolver on Windows. A Bonjour hostname like `JKMBP-M4-Max.local` now resolves on every modern network without explicit configuration; the user can carry their two machines between LANs and the connection still finds them. Drop the `hickory-resolver` dependency entirely; lookup failures surface as `io::Error`, already covered by `ServiceError::Io`.

## 9. Multi-homed DTLS listener (`2c7ce2e` / `4c80ed0`)

When a host has two interfaces on the same subnet (macOS Wi-Fi `en0` + USB-C dock `en7` both on `192.168.1.0/24`), a single `0.0.0.0:port` DTLS listener silently breaks for peers that dial the non-routed IP: the kernel sources its reply from the routing table's preferred interface, so the reply's src-IP doesn't match the 4-tuple the peer expects, and `webrtc-dtls` drops the packet.

Replace the single `0.0.0.0` bind with one `Listener` per local IPv4 address (loopback + link-local skipped), each socket bound to a specific IP so the kernel uses *that* IP as source — symmetric replies guaranteed regardless of the routing table. An `if-watch` supervisor task adds/drops listener slots dynamically on interface up/down; plugging a dock or toggling Wi-Fi no longer requires a lan-mouse restart.

The supervisor also runs a **30-second reconciliation tick** that diffs the live `getifaddrs` set against the listeners HashMap. `if-watch` on macOS uses Network.framework, which doesn't reliably fire `IfEvent::Down` when an interface is administratively disabled (e.g. user toggles Wi-Fi off in System Settings); the polling backup catches whatever the event stream misses, both adds and drops.

Falls back to a single `0.0.0.0` bind only if interface enumeration or every per-IP bind fails — preserves single-NIC behavior and ensures we never silently fail to listen.

Removes the previous user-facing workaround of forcing `ips = ["192.168.1.88"]` on the peer.

## 10. mDNS-SD service-order discovery (`5ae6fee`)

Even with a multi-homed listener (§9), the **dialer** still has to choose which of the peer's IPs to dial first — and plain hostname resolution returns every interface's IP without ranking. `connect_any`'s parallel race picks whichever DTLS handshake completes first, which is RTT-roughly-correct but not always what the user wanted. The classic symptom: Wi-Fi wins the race even when the user has Ethernet ranked higher in macOS's service order, leading to a stuttery session over Wi-Fi while a healthy wired path sits idle.

Each lan-mouse instance now publishes a `_lan-mouse._udp.local.` Bonjour service whose TXT record carries `primary=<ipv4>`, where `<ipv4>` is the IP of the interface that owns the default route — which on macOS reflects service order, on Linux the lowest-metric default route, on Windows whatever `GetBestRoute2` selects. The dialer continuously browses the same service type and caches `peer_hostname → primary_ipv4` in a `Rc<RefCell<HashMap>>` shared with `LanMouseConnection`.

`connect_any` extended with **happy-eyeballs head-start**: if a preferred address is known, dial it alone for 200ms before joining the rest of the candidate list to the race. A healthy preferred path virtually always wins; a broken one only delays connect by 200ms before fallbacks kick in. (Cf. RFC 8305 IPv6→IPv4 fallback delay.)

Subsystem gated by a new `mdns_discovery` config flag (default true) and a corresponding GUI switch under a new "Network Discovery" preferences group. Toggling off unregisters the service, aborts the browse task, and shuts the daemon, but preserves the `primary_cache` so already-known hints stay queryable until overwritten — useful on networks where mDNS multicast (`224.0.0.251`) is firewalled. A 30-second `discovery_refresh_tick` re-publishes the TXT record so it stays accurate when the OS-preferred interface changes (e.g. user toggles Wi-Fi off and Ethernet takes over).

New deps: `mdns-sd` (cross-platform mDNS responder, doesn't piggyback on system Avahi/Bonjour), `netdev` (default-route lookup), `hostname` (local hostname for the service instance name).

Falls back gracefully when `ServiceDaemon::new` fails (multicast group locked / no perms), no interface owns the default route, or the peer isn't announcing (old version or discovery disabled there) — the dialer just sees `preferred = None` and the existing `connect_any` race runs unchanged.

---

## Test plan

Verified locally: macOS host (2056×1329 logical) ↔ Hyprland guest (2400×1500 logical) over both wired and wifi LAN. Crossings at top / middle / bottom of source land proportionally on destination, both directions, including fast back-and-forth within the same second. Wall-press auto-release deadline gate behaves correctly: never fires on healthy crosses, fires after the deadline elapses on locked-peer scenarios. Trackpad and mouse-wheel scrolling forward correctly with both natural and classic preference combinations, including single-notch behavior in Slack (XWayland) and ghostty (native Wayland). macOS Cmd+W / Cmd+Q / GUI singleton verified. Host-lock suppression verified on macOS (`Cmd+Ctrl+Q` lock; cursor refuses to cross until unlock; with cursor already on Linux, push-left wall-press cleanly returns the cursor to the locked Mac). mDNS-SD primary-IP hint verified on a multi-homed Mac (Wi-Fi + Ethernet on same subnet): Linux dialer consistently selects the Ethernet path even when Wi-Fi wins a raw RTT race.

**Per-pair receive-side post-processing** (added later in this branch): verified end-to-end on the same Linux ↔ Mac LAN setup. Toggling Natural Scrolling per-peer inverts forwarded scroll deltas correctly. Mouse Sensitivity slider applies the multiplier on the receiver. Settings persist across multiple cross-back cycles in both directions, confirmed via `peer at <pos> reports sensitivity X` debug logs on both ends. Sub-1.0 sensitivity no longer triggers premature wall-press AutoRelease — the new `ProtoEvent::ReceiverSensitivity` reaches the sender's wall-press model, which now scales its accumulator to match the receiver's actual cursor advance. Live updates: dragging the slider while a cross is active immediately propagates to the other end without waiting for a cross-back-then-cross-forward.

**Peer-version display**: shows green-matched on same-commit pairs and orange-unknown when one side runs a pre-Hello build, and remains correct in one-direction-down scenarios (listen-side mirror). After today's `get_client(active_addr.ip())` fallback and connect-side `Hello`-echo broadcast, the version row also populates regardless of which side dials first, even when DNS resolves to a stale IP that differs from the actual mDNS-primary IP — verified by configuring a peer with mismatched DNS and confirming the version row shows the correct hash on both ends after the first cross.

**mDNS-resolved peer identity in Incoming Connections rows**: subtitle renders as `mac-mini (192.168.1.42)` (or just hostname / just IP / `(not yet connected)` depending on what's known) after first connect, persisted across daemon restart via `last_addr` / `last_hostname` on the fingerprint-keyed peer entry.

**GUI behavior**: toggling per-row settings doesn't collapse the AdwExpanderRow's expansion state. Fingerprint copy button writes to the system clipboard.

**CI**: all 11 PRs in this series (the 9 splits + this one + #437) green at last poll.

Open items (not blocking — environment access):

- [ ] X11, libei, Windows emulation backends for cursor sync + wall-press
- [ ] Sway-on-wlroots (verified on Hyprland)
- [ ] Forward-compat: peer running pre-Bounds build; back-compat: this build with non-Bounds peer
- [ ] Windows host-lock suppression end-to-end (code path mirrors macOS but unverified on a real Windows host)

Build hygiene: `cargo fmt --all` clean, `cargo clippy --workspace --all-targets --all-features -- -D warnings` clean, `cargo test --workspace` passes.

---

## Related PRs

This PR is part of an effort to split #418 into focused, independently-reviewable pieces.

**The split stack** (each builds on the previous; review in order):

1. #420 — wall-press auto-release + Bounds protocol foundation
2. #429 — CursorPos protocol + cursor warps without prior Bounds
3. #430 — host-lock crossing suppression
4. #431 — peer version exchange
5. #432 — multi-homed DTLS listener + OS-resolver DNS
6. #433 — mDNS-SD primary-IP hints
7. #434 — macOS QoL bundle + UI polish
8. #435 — per-pair scroll/sensitivity + scroll & version-exchange fixes
9. #436 — cross-platform GUI singleton

**Plus:**
- #437 — `fix(capture)`: throttle WaitingForAck Enter re-sends to one per 50ms — small standalone fix off `main`, independent of the stack
- #418 — original all-in-one PR; carries everything above. Kept open as a reference.

### DIFF
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,0 +1,1 @@
+{"sessionId":"9ccb244f-2369-4130-b2f1-ad1d8bc6deda","pid":100308,"procStart":"6549470","acquiredAt":1778010488725}
--- a/.claude/scheduled_tasks.lock
+++ b/.claude/scheduled_tasks.lock
@@ -1,1 +1,0 @@
-{"sessionId":"9ccb244f-2369-4130-b2f1-ad1d8bc6deda","pid":100308,"procStart":"6549470","acquiredAt":1778010488725}
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+.claude/
 .gdbinit
 .idea/
 .vs/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-io"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
+dependencies = [
+ "autocfg",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
+ "rustix 1.1.4",
+ "slab",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "async-recursion"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -410,6 +428,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "cipher"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -502,6 +526,16 @@ dependencies = [
 
 [[package]]
 name = "core-foundation"
+version = "0.9.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "91e195e091a93c46f7102ec7818a2aa394e1e1771c3ab4825963fa03e45afb8f"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
+]
+
+[[package]]
+name = "core-foundation"
 version = "0.10.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b2a6cd9ae233e7f62ba4e9353e81a88df7fc8a5987b8d445b4d90c879bd156f6"
@@ -523,7 +557,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "064badf302c3194842cf2c5d61f56cc88e54a759313879cdf03abdd27d0c3b97"
 dependencies = [
  "bitflags 2.11.0",
- "core-foundation",
+ "core-foundation 0.10.1",
  "core-graphics-types",
  "foreign-types",
  "libc",
@@ -536,7 +570,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d44a101f213f6c4cdc1853d4b78aef6db6bdfa3468798cc1d9912f4735013eb"
 dependencies = [
  "bitflags 2.11.0",
- "core-foundation",
+ "core-foundation 0.10.1",
  "libc",
 ]
 
@@ -841,6 +875,12 @@ name = "find-msvc-tools"
 version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5baebc0774151f905a1a2cc41989300b1e6fbb29aff0ceffa1064fdd3088d582"
+
+[[package]]
+name = "fnv"
+version = "1.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
 name = "foldhash"
@@ -1354,6 +1394,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
+name = "hermit-abi"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fc0fef456e4baa96da950455cd02c081ca953b141298e41db3fc7e36b1da849c"
+
+[[package]]
 name = "hex"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1522,6 +1568,49 @@ dependencies = [
 ]
 
 [[package]]
+name = "if-addrs"
+version = "0.13.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "69b2eeee38fef3aa9b4cc5f1beea8a2444fc00e7377cafae396de3f5c2065e24"
+dependencies = [
+ "libc",
+ "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "if-addrs"
+version = "0.15.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c0a05c691e1fae256cf7013d99dad472dc52d5543322761f83ec8d47eab40d2b"
+dependencies = [
+ "libc",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
+name = "if-watch"
+version = "3.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "71c02a5161c313f0cbdbadc511611893584a10a7b6153cb554bdf83ddce99ec2"
+dependencies = [
+ "async-io",
+ "core-foundation 0.9.4",
+ "fnv",
+ "futures",
+ "if-addrs 0.15.0",
+ "ipnet",
+ "log",
+ "netlink-packet-core",
+ "netlink-packet-route",
+ "netlink-proto",
+ "netlink-sys",
+ "rtnetlink",
+ "system-configuration",
+ "tokio",
+ "windows 0.62.2",
+]
+
+[[package]]
 name = "indexmap"
 version = "2.13.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1570,7 +1659,7 @@ dependencies = [
  "ashpd",
  "async-trait",
  "bitflags 2.11.0",
- "core-foundation",
+ "core-foundation 0.10.1",
  "core-foundation-sys",
  "core-graphics",
  "futures",
@@ -1589,7 +1678,7 @@ dependencies = [
  "wayland-client",
  "wayland-protocols",
  "wayland-protocols-wlr",
- "windows",
+ "windows 0.61.3",
  "x11",
 ]
 
@@ -1600,7 +1689,7 @@ dependencies = [
  "ashpd",
  "async-trait",
  "bitflags 2.11.0",
- "core-foundation",
+ "core-foundation 0.10.1",
  "core-foundation-sys",
  "core-graphics",
  "futures",
@@ -1615,7 +1704,7 @@ dependencies = [
  "wayland-protocols",
  "wayland-protocols-misc",
  "wayland-protocols-wlr",
- "windows",
+ "windows 0.61.3",
  "x11",
 ]
 
@@ -1750,6 +1839,8 @@ dependencies = [
  "clap",
  "env_logger",
  "futures",
+ "if-addrs 0.13.4",
+ "if-watch",
  "input-capture",
  "input-emulation",
  "input-event",
@@ -2001,6 +2092,54 @@ dependencies = [
 ]
 
 [[package]]
+name = "netlink-packet-core"
+version = "0.8.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3463cbb78394cb0141e2c926b93fc2197e473394b761986eca3b9da2c63ae0f4"
+dependencies = [
+ "paste",
+]
+
+[[package]]
+name = "netlink-packet-route"
+version = "0.28.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4ce3636fa715e988114552619582b530481fd5ef176a1e5c1bf024077c2c9445"
+dependencies = [
+ "bitflags 2.11.0",
+ "libc",
+ "log",
+ "netlink-packet-core",
+]
+
+[[package]]
+name = "netlink-proto"
+version = "0.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b65d130ee111430e47eed7896ea43ca693c387f097dd97376bffafbf25812128"
+dependencies = [
+ "bytes",
+ "futures",
+ "log",
+ "netlink-packet-core",
+ "netlink-sys",
+ "thiserror 2.0.18",
+]
+
+[[package]]
+name = "netlink-sys"
+version = "0.8.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cd6c30ed10fa69cc491d491b85cc971f6bdeb8e7367b7cde2ee6cc878d583fae"
+dependencies = [
+ "bytes",
+ "futures-util",
+ "libc",
+ "log",
+ "tokio",
+]
+
+[[package]]
 name = "nix"
 version = "0.26.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2011,6 +2150,18 @@ dependencies = [
  "libc",
  "memoffset 0.7.1",
  "pin-utils",
+]
+
+[[package]]
+name = "nix"
+version = "0.30.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "74523f3a35e05aba87a1d978330aef40f67b0304ac79c1c00b294c9830543db6"
+dependencies = [
+ "bitflags 2.11.0",
+ "cfg-if",
+ "cfg_aliases",
+ "libc",
 ]
 
 [[package]]
@@ -2289,6 +2440,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7edddbd0b52d732b21ad9a5fab5c704c14cd949e5e9a1ec5929a24fded1b904c"
 
 [[package]]
+name = "polling"
+version = "3.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5d0e4f59085d47d8241c88ead0f274e8a0cb551f3625263c05eb8dd897c34218"
+dependencies = [
+ "cfg-if",
+ "concurrent-queue",
+ "hermit-abi",
+ "pin-project-lite",
+ "rustix 1.1.4",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "polyval"
 version = "0.6.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2520,6 +2685,24 @@ dependencies = [
  "libc",
  "untrusted",
  "windows-sys 0.52.0",
+]
+
+[[package]]
+name = "rtnetlink"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4b960d5d873a75b5be9761b1e73b146f52dddcd27bac75263f40fba686d4d7b5"
+dependencies = [
+ "futures-channel",
+ "futures-util",
+ "log",
+ "netlink-packet-core",
+ "netlink-packet-route",
+ "netlink-proto",
+ "netlink-sys",
+ "nix 0.30.1",
+ "thiserror 1.0.69",
+ "tokio",
 ]
 
 [[package]]
@@ -2850,6 +3033,27 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "system-configuration"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a13f3d0daba03132c0aa9767f98351b3488edc2c100cda2d2ec2b04f3d8d3c8b"
+dependencies = [
+ "bitflags 2.11.0",
+ "core-foundation 0.9.4",
+ "system-configuration-sys",
+]
+
+[[package]]
+name = "system-configuration-sys"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8e1d1b10ced5ca923a1fcb8d03e96b8d3268065d724548c0211415ff6ac6bac4"
+dependencies = [
+ "core-foundation-sys",
+ "libc",
 ]
 
 [[package]]
@@ -3514,7 +3718,7 @@ dependencies = [
  "lazy_static",
  "libc",
  "log",
- "nix",
+ "nix 0.26.4",
  "portable-atomic",
  "rand",
  "thiserror 1.0.69",
@@ -3559,11 +3763,23 @@ version = "0.61.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9babd3a767a4c1aef6900409f85f5d53ce2544ccdfaa86dad48c91782c6d6893"
 dependencies = [
- "windows-collections",
+ "windows-collections 0.2.0",
  "windows-core 0.61.2",
- "windows-future",
+ "windows-future 0.2.1",
  "windows-link 0.1.3",
- "windows-numerics",
+ "windows-numerics 0.2.0",
+]
+
+[[package]]
+name = "windows"
+version = "0.62.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "527fadee13e0c05939a6a05d5bd6eec6cd2e3dbd648b9f8e447c6518133d8580"
+dependencies = [
+ "windows-collections 0.3.2",
+ "windows-core 0.62.2",
+ "windows-future 0.3.2",
+ "windows-numerics 0.3.1",
 ]
 
 [[package]]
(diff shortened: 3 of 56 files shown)


feschber:
Please take a look at #347 #407, I think we have some overlap here.

As for the first point, I'm a bit skeptic if this works correctly with games / pointer grabbing. I'd also like to post pone this until after 0.11.0 because of the protocol changes.

feschber:
I see, you're doing a lot of work here. It would be great, if you could separate unrelated things into different PRs.
That would make it a lot easier to review.
You don't have to branch of main, we can merge separate PRs in order and then fast forward the others.

jondkinney:
> I see, you're doing a lot of work here. It would be great, if you could separate unrelated things into different PRs. That would make it a lot easier to review. You don't have to branch of main, we can merge separate PRs in order and then fast forward the others.

Hey @feschber - I can try to split things out. I just wanted to fix/add all the things that I wanted to work for me. I just got all the tests passing again, so I'll work on the decomposition next.

feschber:
Thank you! And as I said: No reason to necessarily split everything up. Just take the commits as they are and put them into separate prs :)

jondkinney:
Hey @feschber — quick update on the split. After more iteration the layout is now **9 PRs forming a single linear stack**, plus **one standalone fix**. All 11 PRs are CI-green at the moment.

**Stack** (each builds on the previous; review / merge in order):

1. #420 — wall-press auto-release + `Bounds` protocol foundation
2. #429 — `CursorPos` protocol + cursor warps without prior Bounds
3. #430 — host-lock crossing suppression
4. #431 — peer version exchange (`Hello` proto event)
5. #432 — multi-homed DTLS listener + OS-resolver DNS
6. #433 — mDNS-SD primary-IP hints
7. #434 — macOS QoL bundle + UI polish
8. #435 — per-pair scroll/sensitivity controls + scroll & version-exchange fixes
9. #436 — cross-platform GUI singleton

**Standalone:**

- #437 — `fix(capture): throttle WaitingForAck Enter re-sends to one per 50ms`. Targets `main` directly, independent of the stack — review/merge whenever convenient.

**Merge dance:** each split branch is a strict descendant of the previous, so as you merge in order each subsequent PR's compare URL collapses to only its own commits. No rebase needed from you between merges — just go in order. If you'd rather merge non-sequentially, ping me and I'll rebase the affected branches.

A few things from #347 and #407 were folded in with credit:

- #436 borrows the `is_service_running()` probe and the daemon-revival fallback from #407 (atop a different singleton mechanism — see the comparison table in the PR body).
- #435 borrows the receive-side transform pattern from #347 by @NeoTheFox / Raidon Chrome (originally discussed in #339), with `Co-Authored-By` on the relevant commit. Per-pair scope, GUI surfacing, and the wall-press wire-protocol fix make #435 a superset; per-pair handles the multi-sender case where senders on different OSes emit deltas at incompatible scales that no single global multiplier can satisfy.

I'll leave this PR (#418) open as the umbrella reference — the full architecture write-up and end-to-end test plan are here. Close it whenever you'd like; happy to do it once the splits land.

PR #420, opened 2026-05-06
feat(capture): wall-press auto-release + Bounds protocol foundation

jondkinney:
Wall-press auto-release feature plus the `Bounds` proto event the cursor-sync work builds on.

## Summary
- Cross-platform wall-press auto-release fallback (cursor releases when pushed past a configurable threshold)
- `Bounds(width, height)` proto event for receivers to advertise their display dimensions
- All emulation backends report `display_bounds` and warp the cursor on Enter
- Capture caches peer bounds for use as the wall-press upper clamp

## Test plan
- [ ] Push cursor against the far edge of the peer screen; auto-release fires after threshold
- [ ] Verify on macOS, Windows, and Linux (wlroots/X11)

## Related PRs

This PR is part of an effort to split #418 into focused, independently-reviewable pieces.

**The split stack** (each builds on the previous; review in order):

1. #420 — wall-press auto-release + Bounds protocol foundation
2. #429 — CursorPos protocol + cursor warps without prior Bounds
3. #430 — host-lock crossing suppression
4. #431 — peer version exchange
5. #432 — multi-homed DTLS listener + OS-resolver DNS
6. #433 — mDNS-SD primary-IP hints
7. #434 — macOS QoL bundle + UI polish
8. #435 — per-pair scroll/sensitivity + scroll & version-exchange fixes
9. #436 — cross-platform GUI singleton

**Plus:**
- #437 — `fix(capture)`: throttle WaitingForAck Enter re-sends to one per 50ms — small standalone fix off `main`, independent of the stack
- #418 — original all-in-one PR; carries everything above. Kept open as a reference.

---

CleanShot 2026-05-07 at 09 39 37@2x

### DIFF
--- a/input-capture/src/lib.rs
+++ b/input-capture/src/lib.rs
@@ -9,7 +9,7 @@ use async_trait::async_trait;
 use futures::StreamExt;
 use futures_core::Stream;
 
-use input_event::{Event, KeyboardEvent, scancode};
+use input_event::{Event, KeyboardEvent, PointerEvent, scancode};
 
 pub use error::{CaptureCreationError, CaptureError, InputCaptureError};
 
@@ -41,6 +41,11 @@ pub enum CaptureEvent {
     Begin,
     /// input event coming from capture handle
     Input(Event),
+    /// the capture wrapper detected sustained back-toward-host motion
+    /// past the configured threshold (the user has pinned the cursor
+    /// at the host-adjacent edge of the guest and kept pushing). The
+    /// capture loop should treat this like a release-bind chord.
+    AutoRelease,
 }
 
 impl Display for CaptureEvent {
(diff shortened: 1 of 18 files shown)


jondkinney:
@feschber this is the new base of the work that I split out. Sorry for the noise of closing the last stack of them. That happened automatically when I was trying to rename the branches to be in a good review order and be as atomic as possible.

This particular PR is made against main, but starting with #429 you can review each of the PRs against the prior using the `Review-only focused diff (just this PR's commits, vs.` note at the top of the description of each PR, which shows a separate comparison against the prior PR.

feschber:
I'm not particularly in favor of this change...

The fundamental problem is that we CAN NOT track where the cursor is from the client side alone because of
1. pointer acceleration
2. pointer grabbing

Currently not all backends actually emulate physical input events but my goal is to achieve this at some point.
Having phyiscal input emulation makes it impossible to track the cursor position client side because of the reasons stated above.

PR #421, opened 2026-05-06
feat: peer version exchange via Hello proto event

jondkinney:
**Review-only focused diff** (just this PR's commits, vs. `split/01-cursor-and-walls`): https://github.com/jondkinney/lan-mouse/compare/split/01-host-lock...split/02-peer-version

## Summary

Peer version exchange — each peer surfaces the other's build commit hash in the GUI, with a soft-warn color indicator on mismatch.

`ProtoEvent::Hello { commit: [u8; 8] }` carries each peer's `shadow_rs` `SHORT_COMMIT` once per session. Sender fires immediately after DTLS auth; listener mirrors the event back so the connect side's `receive_loop` populates `ClientState::peer_commit` for the right handle. Disconnect path clears it.

Each outgoing-connection row's collapsed subtitle renders match status with Pango-colored markup: green when commits match, orange when mismatched or when the peer hasn't sent `Hello` (older build). Soft-warn only — version mismatch never refuses traffic. The local commit reaches the GTK frontend (separate process) via an explicit `local_commit` parameter on `lan_mouse_gtk::run`, stashed in a `OnceLock` so per-row UI can compare against each peer's hash without an IPC round-trip.

`EventType::Hello` is appended to the enum so existing IDs are untouched. Old peers hit the existing `InvalidEventId` skip path from `3025422` and silently ignore the event — backward interop preserved.

**Listen-side mirror (`1ea7148`):** the original implementation read `peer_commit` only off the outgoing-connect path, on the assumption that bidirectional setups always have a working outbound connection in both directions. That assumption broke the moment any direction's outbound was down (e.g. peer's TCP listener temporarily not bound) — version display silently said "unknown" while the peer was happily sending events to us inbound. New `EmulationEvent::PeerHello { addr, commit }` variant fired from the listen-side Hello handler; service maps `addr → ClientHandle` via `client_manager.get_client(addr)` and stamps `peer_commit` exactly like the connect path. Version visibility is now independent of outbound reachability.

## Test plan

- [x] Same-commit pairs render the row's subtitle with a green "matched" indicator
- [x] Cross-build pairs (one side pre-Hello) render orange "unknown" — verified bidirectionally
- [x] One-direction-down setups (peer's outbound to us is broken) still show the peer's version because of the listen-side mirror
- [x] Disconnecting a peer clears `peer_commit` so the next connect starts fresh

---

Split out from #418, the umbrella PR collecting ~10 independent feature areas. This PR is the peer-version-exchange subset and stacks on top of the cursor-sync/wall-press PR. See #418 for the full picture.

## Stack overview

6. #425 — scroll forwarding
7. #426 — GUI singleton

### DIFF
--- a/input-capture/Cargo.toml
+++ b/input-capture/Cargo.toml
@@ -58,6 +58,7 @@ bitflags = "2.6.0"
 [target.'cfg(windows)'.dependencies]
 windows = { version = "0.61.2", features = [
     "Win32_System_LibraryLoader",
+    "Win32_System_RemoteDesktop",
     "Win32_System_Threading",
     "Win32_Foundation",
     "Win32_Graphics",
--- a/input-capture/src/dummy.rs
+++ b/input-capture/src/dummy.rs
@@ -42,7 +42,7 @@ impl Capture for DummyInputCapture {
         Ok(())
     }
 
-    async fn release(&mut self) -> Result<(), CaptureError> {
+    async fn release(&mut self, _warp_target: Option<(i32, i32)>) -> Result<(), CaptureError> {
         Ok(())
     }
 
--- a/input-capture/src/dummy.rs
+++ b/input-capture/src/dummy.rs
@@ -62,7 +62,7 @@ impl Stream for DummyInputCapture {
         let event = match self.start {
             None => {
                 self.start.replace(current);
-                CaptureEvent::Begin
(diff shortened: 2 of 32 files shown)


jondkinney:
Superseded by #431 (same content; the branch was renumbered as part of restructuring the stack from 7 PRs into 9, which auto-closed this cross-fork PR).

PR #422, opened 2026-05-06
fix: hostname resolution via OS resolver + multi-homed DTLS listener

jondkinney:
**Review-only focused diff** (just this PR's commits, vs. `split/02-peer-version`): https://github.com/jondkinney/lan-mouse/compare/split/02-peer-version...split/03-network

## Summary

Two unrelated-but-bundled fixes for "lan-mouse should Just Work on real LANs" — one for hostname resolution, one for multi-homed hosts. Together they remove the most common need for the manual `ips=[…]` workaround.

## Hostname resolution via the OS resolver (`9ce3847`)

`hickory_resolver::TokioResolver` only consults `/etc/resolv.conf` and queries upstream DNS servers — which means it can't see `/etc/hosts`, mDNS/Avahi/Bonjour, NetBIOS, or anything else in the system's full name-resolution stack. On a typical home LAN there's no DNS server that knows about peer machine names, so users had to fall back to typing IP addresses, which broke the moment they moved their setup to a different network.

Swap to `tokio::net::lookup_host`, which calls `getaddrinfo`. That walks `/etc/nsswitch.conf` on Linux (picking up Avahi-resolved `.local` names, `/etc/hosts`, and DNS), uses Bonjour for `.local` on macOS, and the full Windows resolver on Windows. A Bonjour hostname like `JKMBP-M4-Max.local` now resolves on every modern network without explicit configuration; the user can carry their two machines between LANs and the connection still finds them. Drop the `hickory-resolver` dependency entirely; lookup failures surface as `io::Error`, already covered by `ServiceError::Io`.

## Multi-homed DTLS listener (`2c7ce2e` / `4c80ed0`)

When a host has two interfaces on the same subnet (macOS Wi-Fi `en0` + USB-C dock `en7` both on `192.168.1.0/24`), a single `0.0.0.0:port` DTLS listener silently breaks for peers that dial the non-routed IP: the kernel sources its reply from the routing table's preferred interface, so the reply's src-IP doesn't match the 4-tuple the peer expects, and `webrtc-dtls` drops the packet.

Replace the single `0.0.0.0` bind with one `Listener` per local IPv4 address (loopback + link-local skipped), each socket bound to a specific IP so the kernel uses *that* IP as source — symmetric replies guaranteed regardless of the routing table. An `if-watch` supervisor task adds/drops listener slots dynamically on interface up/down; plugging a dock or toggling Wi-Fi no longer requires a lan-mouse restart.

The supervisor also runs a **30-second reconciliation tick** that diffs the live `getifaddrs` set against the listeners HashMap. `if-watch` on macOS uses Network.framework, which doesn't reliably fire `IfEvent::Down` when an interface is administratively disabled (e.g. user toggles Wi-Fi off in System Settings); the polling backup catches whatever the event stream misses, both adds and drops.

Falls back to a single `0.0.0.0` bind only if interface enumeration or every per-IP bind fails — preserves single-NIC behavior and ensures we never silently fail to listen.

Removes the previous user-facing workaround of forcing `ips = ["192.168.1.88"]` on the peer.

## Test plan

- [x] `JKMBP-M4-Max.local` resolves on a fresh network without any DNS-server config
- [x] Multi-homed Mac (Wi-Fi + Ethernet on same subnet): peer can dial either IP and the DTLS handshake completes
- [x] Toggling Wi-Fi off in macOS System Settings: the listener slot drops within ~30s (reconciliation tick) without restart
- [x] Plugging in a USB-C dock: the new IP gets a listener slot via `if-watch`
- [x] Single-NIC machines fall back to the existing `0.0.0.0` bind (no behavior change)

---

Split out from #418, the umbrella PR collecting ~10 independent feature areas. This PR contains the hostname-resolution and multi-homed-listener subsets. See #418 for the full picture.

## Stack overview

6. #425 — scroll forwarding
7. #426 — GUI singleton

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-io"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
+dependencies = [
+ "autocfg",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
+ "rustix 1.1.4",
+ "slab",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "async-recursion"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -410,6 +428,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "cipher"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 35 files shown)


feschber:
The dns lookup changes seem reasonable!
However, I don't quite understand why we need one listener per interface?
What problem does this solve?
From my understanding this worked because a client trying to connect would just try all exposed IPs (queried from DNS) and the first one to reply would then be set as the default for the client until it becomes unreachable.
What exactly do you mean by silently breaking?

jondkinney:
> What exactly do you mean by silently breaking?

It would fail to cross over/work at all.

jondkinney:
Superseded by #432 (same content; the branch was renumbered as part of restructuring the stack from 7 PRs into 9, which auto-closed this cross-fork PR).

PR #424, opened 2026-05-06
macOS: QoL bundle (LSUIElement, TCC flow, quit-unfreezable, display wake) + UI polish

jondkinney:
**Review-only focused diff** (just this PR's commits, vs. `split/04-mdns`): https://github.com/jondkinney/lan-mouse/compare/split/04-mdns...split/05-macos-qol

## Summary

Quality-of-life improvements primarily targeting the macOS bundle, plus generally-applicable UI polish that touches the same files.

### macOS-specific

- **Cmd+W (`afe9456`)** wires to GtkWindow's `window.close` action, which on macOS hides the window and flips the activation policy to `Accessory` — effectively collapses to the menu bar. Linux/Windows unchanged.
- **Default height 700→1400 (`8a86b9d`)** fits every preference group on first paint on ≥1440px-tall displays. Tiling Wayland WMs ignore `default-height`, so no effect there.
- **Quit unfreezable (`39752ee`)**: 3s `try_wait()` poll on the daemon child with SIGKILL fallback, plus a process-level `std::thread` 5s force-exit backstop scheduled *outside* the GTK main loop (so a wedged loop can't prevent it). Worst-case quit latency 5s; normal completes in <1s.
- **Display wake on input (`56f828f`)**: emulation backend pokes `IOPMAssertionDeclareUserActivity` whenever a forwarded event arrives, so a peer-driven keystroke or click wakes the macOS display from idle-sleep. Without this, lan-mouse would keep pumping events to a blanked screen.
- **TCC.db watcher with fresh-subprocess AX probe (`5aa3fcc`)**: detects "remove from list" cases where `AXIsProcessTrusted` in the parent process keeps reporting cached-true even after the entry is removed. Triggers a relaunch prompt rather than letting the daemon silently lose Accessibility.
- **Quit on AX revoked mid-session (`9708d9c`)**: an active CGEventTap at HeadInsertEventTap can wedge system input if the process lingers after losing AX, so we exit immediately.
- **Re-present window on app re-launch (`53c668b`)**: register the existing status-item delegate as a direct handler for `'aevt'/'rapp'` via NSAppleEventManager so reopening the app via Finder/Dock/`open` brings the window back.
- **Various TCC / Accessibility flow polish**: per-pane navigation, Sequoia-tolerant permission flow, hide warning row once granted, fold relaunch prompt into the warning row.

### Generally-applicable UI polish

- **Wrap window content in `GtkScrolledWindow`** with `propagate-natural-height` — the window can shrink under tiling without clipping content.
- **Modal dialogs** (Add Certificate, etc.): cap width at 460px and size to `parent − 40`, replace `AdwPreferencesGroup` wrappers with plain `GtkEntry`, tighten vertical spacing, close on Escape.
- **Title-case section labels, period-terminated subtitles, vertical-rhythm tightening, hostname:port collapsed row** — cosmetic but consistent.
- **GUI quits when the daemon's IPC connection drops (`7fcc08e`)**: prevents a zombie window after the daemon exits.

## Test plan

- [x] Cmd+W collapses the macOS app to the menu bar; the menubar item still works
- [x] Default height of 1400px shows all preference groups without scrolling on a 1440px-tall display
- [x] Quit completes in <1s normally; SIGKILL fires within 3s when the daemon is wedged; process-level backstop fires within 5s if the GTK loop itself is wedged
- [x] Forwarded keystroke from a peer wakes the macOS display from idle sleep
- [x] Revoking Accessibility mid-session causes the app to exit immediately (no zombie tap)
- [x] Reopening the app via Finder while it's running brings the window forward (no duplicate process)

---

Split out from #418, the umbrella PR collecting ~10 independent feature areas. This PR is the macOS QoL + UI polish subset. See #418 for the full picture.

## Stack overview

6. #425 — scroll forwarding
7. #426 — GUI singleton

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+.claude/
 .gdbinit
 .idea/
 .vs/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-io"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
+dependencies = [
+ "autocfg",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
+ "rustix 1.1.4",
+ "slab",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "async-recursion"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -410,6 +428,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
+name = "cfg_aliases"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "613afe47fcd5fac7ccf1db93babcb082c5994d996f20b8b159f2ad1658eb5724"
+
+[[package]]
 name = "cipher"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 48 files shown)


jondkinney:
Superseded by #434 (same content; the branch was renumbered as part of restructuring the stack from 7 PRs into 9, which auto-closed this cross-fork PR).

PR #425, opened 2026-05-06
feat(scroll): receiver natural-scroll toggle + wlroots axis_source + macOS v120 fix

jondkinney:
**Review-only focused diff** (just this PR's commits, vs. `split/05-macos-qol`): https://github.com/jondkinney/lan-mouse/compare/split/05-macos-qol...split/06-scroll

## Summary

Four scroll-handling fixes that together make trackpad and mouse-wheel forwarding behave correctly across natural/classic preference combinations.

**macOS capture (`dc09e88`):** Wire convention canonicalized to classic mouse-wheel direction regardless of the user's macOS Natural Scrolling preference (read via `CFPreferencesCopyAppValue`). Without a fixed wire convention, the receiver-side toggle has no stable reference. Opposite of #415, which negated unconditionally and traded the bug for the opposite-preference user.

**wlroots emulation (`a26ae8b`):** Continuous-scroll events (trackpad) were silently dropped by Hyprland/Sway/GNOME-Shell because the `axis()` event lacked a companion `axis_source` in the same `wl_pointer.frame`. Spec calls it a "hint"; in practice it's load-bearing for continuous scroll. One extra emit per axis frame fixes it.

**Receiver-side `Natural scrolling` toggle (`9ed25b0`):** New GUI preference, off by default — wire is classic, receiver re-inverts. Mirrors libinput's `natural_scroll`, applied to forwarded events specifically (which on Wayland bypass libinput).

**macOS line→tick mapping (`19d36bf`):** Capture used to map one macOS scroll-line to 40 v120 units (1/3 of a wheel tick), so receivers using the discrete count (Slack via XWayland, terminals reading `axis_value120`) needed 3+ notches before any scroll registered. macOS already amplifies `SCROLL_WHEEL_EVENT_DELTA` by velocity, so one line should map directly to one full v120 tick. Slow-notch and fast-flick behavior is now symmetric across native-Wayland and XWayland clients.

## Test plan

- [x] Trackpad scroll on macOS host reaches Hyprland peer (was previously silently dropped on continuous-scroll path)
- [x] Mouse wheel notch on macOS produces exactly one `axis_value120 = 120` event on the wire
- [x] Slack-via-XWayland and ghostty-on-native-Wayland both register a single notch as a single scroll on the peer
- [x] Receiver-side natural-scroll toggle inverts forwarded events; physical input on the receiver continues to use libinput's preference
- [x] Both natural-on-sender × natural-on-receiver and the cross-product cases produce intuitive direction

---

Split out from #418, the umbrella PR collecting ~10 independent feature areas. This PR is the scroll-handling subset. See #418 for the full picture.

## Stack overview

6. #425 — scroll forwarding
7. #426 — GUI singleton

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+.claude/
 .gdbinit
 .idea/
 .vs/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-io"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
+dependencies = [
+ "autocfg",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
+ "rustix 1.1.4",
+ "slab",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "async-recursion"
 version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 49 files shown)


jondkinney:
Superseded by #435 (same content; the branch was renumbered as part of restructuring the stack from 7 PRs into 9, which auto-closed this cross-fork PR).

PR #426, opened 2026-05-06
feat(gui): cross-platform GUI singleton via dedicated socket

jondkinney:
**Review-only focused diff** (just this PR's commits, vs. `split/06-scroll`): https://github.com/jondkinney/lan-mouse/compare/split/06-scroll...split/07-singleton

## Summary

Cross-platform GUI singleton — `lan-mouse` launched twice no longer opens a second window.

Dedicated `lan-mouse-gui.sock` (Unix) / `127.0.0.1:5253` (Windows), separate from the daemon socket. First GUI binds; later launches connect, send a byte, and exit. Primary forwards into the GTK main loop and calls `window.present()`.

Decoupled from the daemon socket on purpose: `lan-mouse daemon` headless workflow is unchanged. Stale-socket recovery if the primary crashed without cleanup. Defense-in-depth via `app.windows().first()` for the in-process activation path. Unit test covers acquire → signal → re-acquire.

## Test plan

- [x] First GUI invocation binds the singleton socket; window opens
- [x] Second invocation while the first is running: the second exits cleanly and the first's window is brought forward
- [x] Primary crashed without cleanup → next GUI invocation reclaims the stale socket and starts fresh
- [x] Headless `lan-mouse daemon` workflow unchanged (singleton lock only exists while a GUI is running)
- [x] In-process activation (DBus single-instance on Linux, `kAEReopenApplication` on macOS) reuses the existing window via `app.windows().first()`

---

Split out from #418, the umbrella PR collecting ~10 independent feature areas. This PR is the GUI singleton subset. See #418 for the full picture.

## Stack overview

6. #425 — scroll forwarding
7. #426 — GUI singleton

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+.claude/
 .gdbinit
 .idea/
 .vs/
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -202,6 +202,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-io"
+version = "2.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "456b8a8feb6f42d237746d4b3e9a178494627745c3c56c6ea55d92ba50d026fc"
+dependencies = [
+ "autocfg",
+ "cfg-if",
+ "concurrent-queue",
+ "futures-io",
+ "futures-lite",
+ "parking",
+ "polling",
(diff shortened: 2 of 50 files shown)


jondkinney:
Superseded by #436 (same content; the branch was renumbered as part of restructuring the stack from 7 PRs into 9, which auto-closed this cross-fork PR).

PR #445, opened 2026-05-22
macos: map Print Screen / Scroll Lock / Pause to F13/F14/F15

tyvsmith:
## Problem

The `keycode` crate's Chromium-derived evdev→CGKeyCode table has no entry for three keys that appear on virtually every standard PC keyboard:

- `KEY_SYSRQ` (99) — Print Screen
- `KEY_SCROLLLOCK` (70) — Scroll Lock
- `KEY_PAUSE` (119) — Pause / Break

When a Linux client sends any of these keys, the macOS emulation backend hits the `Err(_)` arm and silently drops the event with `"unable to map key event"`.

This is reproducible with any software KVM (e.g. lan-mouse itself) where the sending machine is Linux and the receiving machine is macOS: pressing Print Screen on the Linux keyboard produces `keycode: 0xffff` on the macOS side with no key event emitted.

## Fix

Add a fallback match on `scancode::Linux` before delegating to `KeyMap::from_key_mapping`. Maps the three keys to F13/F14/F15 (`0x69`/`0x6B`/`0x71`) respectively — the decades-old macOS convention for PC keys that have no Apple keyboard equivalent.

This allows tools on the macOS side (e.g. `skhd`) to bind these keys normally, since F13/F14/F15 are well-known CGKeyCodes.

## Testing

- Sent Print Screen, Scroll Lock, and Pause from an Arch Linux desktop to macOS via lan-mouse
- Confirmed `skhd -o` reports `keycode: 0x69` / `0x6B` / `0x71` for each key respectively (previously `0xffff`)
- All other key events unaffected (fallback to existing `KeyMap::from_key_mapping` path)

---

> **Reparented on #449.** This branch is temporarily stacked on top of #449 (`fix(ci): bust stale Homebrew glib cache on macOS runners`). The `build macos-latest` / `test macos-latest` failures here were **not** caused by this change — they are the stale Homebrew `glib` cache regression (`search path '/opt/homebrew/Cellar/glib/2.88.0/lib' not found` → `library 'gio-2.0' not found`) that affects all open PRs and is fixed by #449. Stacking on #449 makes CI green here now. Once #449 merges, this PR will merge cleanly — its CI-fix commit drops out as a no-op after a rebase onto the updated `main`. The only functional change in this PR remains the macOS keycode mapping in `input-emulation/src/macos.rs`.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -83,7 +83,7 @@ jobs:
           py -m venv .venv
           .venv\Scripts\activate.ps1
           py -m pip install gvsbuild
-          gvsbuild build --msys-dir=C:\msys64 gtk4 libadwaita librsvg
+          gvsbuild build --vs-ver vs2026 --msys-dir=C:\msys64 gtk4 libadwaita librsvg
       - name: cargo build
         if: matrix.job == 'build'
         run: cargo build
--- a/input-emulation/src/macos.rs
+++ b/input-emulation/src/macos.rs
@@ -484,12 +484,21 @@ impl Emulation for MacOSEmulation {
                     key,
                     state,
                 } => {
-                    let code = match KeyMap::from_key_mapping(KeyMapping::Evdev(key as u16)) {
-                        Ok(k) => k.mac as CGKeyCode,
-                        Err(_) => {
-                            log::warn!("unable to map key event");
-                            return Ok(());
-                        }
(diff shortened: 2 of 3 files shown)


tyvsmith:
This is now reparented on #449, so once that merges on main, this will merge cleanly without the seemingly duplicated file to the github workflow.

PR #452, opened 2026-06-07
fix(macos): undo natural scrolling inversion in captured scroll events

### LINKED ISSUE #451: macOS: Natural Scrolling inverts scroll direction for remote clients
## Description

When Natural Scrolling is enabled on the macOS sending side (the default on MacBooks), captured scroll events are already inverted by the system before reaching the CGEvent tap. This causes remote clients to receive scroll deltas in the wrong direction — content-movement direction instead of device/wheel direction.

## Expected Behavior

The network protocol should always carry device-direction (hardware wheel rotation) deltas regardless of the sender's OS scroll preferences. Each receiving OS can then apply its own scroll preference independently.

## Actual Behavior

With Natural Scrolling enabled on macOS, scrolling on the remote machine is inverted compared to what the user expects.

## Proposed Fix

Check CGEvent field 137 (`isDirectionInvertedFromDevice`, corresponding to `NSEvent.isDirectionInvertedFromDevice`) in the macOS input capture backend. When the field is non-zero (system has inverted the scroll direction), negate both continuous and discrete scroll deltas before sending them over the network.

This field is defined in Apple's `CGEventTypes.h` and is widely relied upon by established macOS input utilities (mac-mouse-fix, Scroll Reverser, IINA, etc.).

## Environment

- macOS (tested on Apple Silicon)
- Natural Scrolling enabled (System Settings → Trackpad → Scroll & Zoom)
- Affects both continuous (trackpad) and discrete (mouse wheel) scroll events

Oseenix:
Closes #451

## Problem

When Natural Scrolling is enabled on the macOS sending side (the default on MacBooks), the system inverts scroll deltas before they reach the CGEvent tap. This causes remote clients to receive content-movement direction instead of device/wheel direction, resulting in inverted scrolling on the receiving end.

## Fix

Check CGEvent field 137 (`isDirectionInvertedFromDevice`) — non-zero when the system has inverted scroll direction. When set, negate both continuous (trackpad) and discrete (mouse wheel) scroll deltas before sending them over the network.

This ensures the protocol always carries device-direction deltas, allowing each receiving OS to apply its own scroll preference independently.

Field 137 is defined in Apple's `CGEventTypes.h` and is widely used by established macOS input utilities (mac-mouse-fix, Scroll Reverser, IINA, etc.).

## Changes

- `input-capture/src/macos.rs`: Read `DIRECTION_INVERTED_FROM_DEVICE` (field 137) on scroll events and negate deltas when the system has inverted them. Applies to both continuous (`SCROLL_WHEEL_EVENT_POINT_DELTA_AXIS_{1,2}`) and discrete (`SCROLL_WHEEL_EVENT_DELTA_AXIS_{1,2}`) scroll paths.

### DIFF
--- a/input-capture/src/macos.rs
+++ b/input-capture/src/macos.rs
@@ -12,7 +12,7 @@ use core_graphics::{
     base::{CGError, kCGErrorSuccess},
     display::{CGDisplay, CGPoint},
     event::{
-        CGEvent, CGEventFlags, CGEventTap, CGEventTapLocation, CGEventTapOptions,
+        CGEvent, CGEventField, CGEventFlags, CGEventTap, CGEventTapLocation, CGEventTapOptions,
         CGEventTapPlacement, CGEventTapProxy, CGEventType, CallbackResult, EventField,
     },
     event_source::{CGEventSource, CGEventSourceStateID},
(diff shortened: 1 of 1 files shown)


Oseenix:
Closing: need more testing to verify the fix is reliable.

PR #453, opened 2026-06-07
fix(macos): correct scroll direction for remote clients

### LINKED ISSUE #451: macOS: Natural Scrolling inverts scroll direction for remote clients

Oseenix:
Closes #451

## Problem

macOS scroll deltas use the opposite sign convention from evdev when Natural Scrolling is OFF. When Natural Scrolling is ON, the system negates the deltas, which happens to match evdev. This causes remote Linux/Windows clients to receive inverted scroll events depending on the macOS sender's scroll preference.

## Fix

Read CGEvent field 137 (`isDirectionInvertedFromDevice`) to detect whether Natural Scrolling is active. When it is not, negate both continuous (trackpad) and discrete (mouse wheel) scroll deltas so the network protocol always carries evdev-compatible values.

Field 137 corresponds to `NSEvent.isDirectionInvertedFromDevice` and is defined in Apple's `CGEventTypes.h`.

## Testing

Verified on macOS (Apple Silicon) with both Natural Scrolling ON and OFF — scroll direction on Linux remote client is now correct in both cases.

## Changes

- `input-capture/src/macos.rs`: Check `isDirectionInvertedFromDevice` and conditionally negate scroll deltas for both continuous and discrete scroll paths.

### DIFF
--- a/input-capture/src/macos.rs
+++ b/input-capture/src/macos.rs
@@ -12,7 +12,7 @@ use core_graphics::{
     base::{CGError, kCGErrorSuccess},
     display::{CGDisplay, CGPoint},
     event::{
-        CGEvent, CGEventFlags, CGEventTap, CGEventTapLocation, CGEventTapOptions,
+        CGEvent, CGEventField, CGEventFlags, CGEventTap, CGEventTapLocation, CGEventTapOptions,
         CGEventTapPlacement, CGEventTapProxy, CGEventType, CallbackResult, EventField,
     },
     event_source::{CGEventSource, CGEventSourceStateID},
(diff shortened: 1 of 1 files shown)


grazzolini:
Uh, this is what #435 does?

Oseenix:
> Uh, this is what #435 does?
It seems so, the #435 plan looks more complete.

PR #458, opened 2026-06-14
Peer version exchange

feschber:


feschber:
@jondkinney this is definitely a good change, I rebased it on top of main and fixed the styling for light mode

PR #468, opened 2026-07-16
fix(macos): Handle @rpath references in dylib bundling script

dodidev:
## Summary

**Fix macOS runtime crash caused by missing transitive `@rpath` dependencies.**

The `copy-macos-dylib.sh` script previously only copied dynamic libraries with absolute Homebrew paths. This caused runtime crashes (e.g., `dyld: library not loaded`) because it missed transitive dependencies that rely on relative `@rpath` references—such as `libwebp.7.dylib` failing to load `libsharpyuv.0.dylib`.

This PR enhances the dependency resolution inside `scripts/copy-macos-dylib.sh` to recursively discover, copy, and link both absolute and `@rpath` Homebrew libraries.

## Key Changes

* **Dynamic `@rpath` Extraction**: Uses `otool -L` to scan binaries and discover all nested `@rpath` dependencies.
* **Recursive Dependency Resolution**: Locates the missing `@rpath` libraries within Homebrew, copies them to the app bundle, and updates their install names.
* **Error Prevention**: Emits clear warnings if an `@rpath` library cannot be found in the expected Homebrew path to simplify troubleshooting.

## Evidence / Error Log

```sh
dyld[17093]: library not loaded: @rpath/libsharpyuv.0.dylib
  referenced from: <d5d18de0-5439-3b6e-8118-b5d7b486ec66> /applications/lan mouse.app/contents/frameworks/libwebp.7.dylib
  reason: tried: 
    '/applications/lan mouse.app/contents/frameworks/../lib/libsharpyuv.0.dylib' (no such file), 
    '/applications/lan mouse.app/contents/frameworks/libsharpyuv.0.dylib' (no such file), 
    '/applications/lan mouse.app/contents/frameworks/libsharpyuv.0.dylib' (no such file)
zsh: abort      /applications/lan\ mouse.app/contents/macos/lan-mouse

### DIFF
--- a/scripts/copy-macos-dylib.sh
+++ b/scripts/copy-macos-dylib.sh
@@ -57,8 +57,11 @@ share_path="$resources_path/share"
 fix_references() {
   local bin="$1"
 
-  # Get all Homebrew libraries referenced by the binary
+  # Get all Homebrew libraries referenced by the binary (absolute paths)
   libs=$(otool -L "$bin" | awk -v homebrew="$homebrew_path" '$0 ~ homebrew {print $1}')
+
+  # Also get @rpath references and try to resolve them from Homebrew
+  rpath_libs=$(otool -L "$bin" | awk '$1 ~ /^@rpath\// {print $1}')
 
   echo "$libs" | while IFS= read -r old_path; do
     if [ -z "$old_path" ]; then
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by magicAyyub** [comment]
I can confirm this issue on macOS Apple Silicon (arm64) with Lan Mouse 0.11.0 (20260831.090146).

I downloaded lan-mouse-macos-arm64.zip from the latest release. The app would not start because libsharpyuv.0.dylib is missing from the application bundle.

Running the binary directly produced:

```bash
dyld: Library not loaded: @rpath/libsharpyuv.0.dylib
Referenced from: .../Lan Mouse.app/Contents/Frameworks/libwebp.7.dylib
Reason: tried: .../Lan Mouse.app/Contents/Frameworks/libsharpyuv.0.dylib (no such file)
```

libwebp.7.dylib itself is present and is arm64, but libsharpyuv.0.dylib is missing from Contents/Frameworks/.

As a workaround, I copied the matching library from Homebrew:

```BASH
cp /opt/homebrew/opt/webp/lib/libsharpyuv.0.dylib \
"/Applications/Lan Mouse.app/Contents/Frameworks/"
```

After removing the quarantine attribute:

xattr -dr com.apple.quarantine "/Applications/Lan Mouse.app"

Lan Mouse starts and works correctly.

So this appears to be a packaging/bundling issue where the @rpath dependency of libwebp.7.dylib is not included in the macOS application bundle.

System: macOS, Apple Silicon (arm64)
Lan Mouse: 0.11.0
Build: 20260831.090146

The workaround above works for me.

feschber:
Maybe at this point its worth using some existing tool for this?
https://github.com/auriamg/macdylibbundler

magicAyyub:
> Maybe at this point its worth using some existing tool for this? https://github.com/auriamg/macdylibbundler

interesting, thank you.

feschber:
If you want, you could integrate that

feschber:
I decided to review the script using claude. It found a few more issues.
Now everything should be handled correctly as far as I can tell...

PR #471, opened 2026-07-25
Add configurable input post-processing

mrcha033:
## Summary

- add optional `[input_post_processing]` settings for pointer sensitivity and scroll inversion
- apply configuration changes immediately to active input emulation, including CLI-triggered updates
- add `set-mouse-sensitivity` and `invert-scrolling` CLI commands
- document the settings and cover event transformation and CLI parsing with unit tests

If the new configuration section is omitted, behavior remains unchanged (`mouse_sensitivity = 1.0`, `invert_scroll = false`).

## Motivation

Pointer speed and scroll direction can feel inconsistent when input crosses operating systems. Receiver-side post-processing lets users compensate without changing the local mouse settings on either machine.

## Validation

- `cargo clippy --workspace --all-targets --all-features --locked -- -D warnings`
- manually verified macOS-to-Arch/Hyprland pointer sensitivity adjustment

## Attribution

This refreshes and supersedes #347 by @NeoTheFox, rebasing the work onto current `main` and adding runtime propagation, documentation, and tests. The implementation commit retains the original author as a co-author.

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,13 @@ release_bind = [ "KeyA", "KeyS", "KeyD", "KeyF" ]
 # optional port (defaults to 4242)
 port = 4242
 
+# optional transformations applied to input received from other devices
+[input_post_processing]
+# scale relative pointer motion (defaults to 1.0)
+mouse_sensitivity = 1.0
+# invert continuous and discrete scrolling (defaults to false)
+invert_scroll = false
+
 # list of authorized tls certificate fingerprints that
 # are accepted for incoming traffic
 [authorized_fingerprints]
@@ -414,6 +421,13 @@ port = 4242
 ```
 
 Where `left` can be either `left`, `right`, `top` or `bottom`.
+Input post-processing is configured on the receiving device and applies only to
+events emulated there. Both options can also be changed at runtime:
+
+```sh
+lan-mouse cli set-mouse-sensitivity 1.5
+lan-mouse cli invert-scrolling true
+```
 
 ## Roadmap
 - [x] Graphical frontend (gtk + libadwaita)
--- a/config.toml
+++ b/config.toml
@@ -1,10 +1,16 @@
 # example configuration
 
 # configure release bind
-release_bind = [ "KeyA", "KeyS", "KeyD", "KeyF" ]
+release_bind = ["KeyA", "KeyS", "KeyD", "KeyF"]
 
 # optional port (defaults to 4242)
 port = 4242
+
+[input_post_processing]
+# scale relative pointer motion received from other devices
+mouse_sensitivity = 1.0
+# invert scroll events received from other devices
+invert_scroll = false
 
 # list of authorized tls certificate fingerprints that
 # are accepted for incoming traffic
--- a/input-emulation/src/lib.rs
+++ b/input-emulation/src/lib.rs
@@ -4,7 +4,7 @@ use std::{
     fmt::Display,
 };
 
-use input_event::{Event, KeyboardEvent};
+use input_event::{Event, KeyboardEvent, PointerEvent};
 
 pub use self::error::{EmulationCreationError, EmulationError, InputEmulationError};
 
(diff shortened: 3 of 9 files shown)


10a5er:
sorry for asking, its already nice work:

do you see a possebility to have a scroll-speed value?

> when you use something like sway on a macbook, or other exotic input-device/software combination, you sometimes are forced you use very untraditional values for scroll-speed globally for a trackpad to work, which is fine until two (that i encountered) cases:
> 
>   1. controlling a computer over vnc
>   2. lan-mouse
> 
> in my case, we are talking about scrolling factor is reduces to 5-10%

i can imagine there are local solutions and that many settings if such a slider would exist could lead to bad scrolling-resolution, but i do think if its not that unrealistic to implement the simple existance of the slider would not hurt anyone. :D

oscarnogueira:
Tested this PR and it works well for my setup.

  **Setup**
  - Sender: macOS, lan-mouse , mouse with a scroll wheel
  - Receiver: Arch Linux (Omarchy), Hyprland 0.56.2, `wlroots` emulation backend
  - Receiver built from this branch (`83e66bc2`), release profile

  **Config on the receiver**
  ```toml
  [input_post_processing]
  invert_scroll = true

  Results
  - Without this PR, mouse wheel scrolling from the Mac arrives inverted on Hyprland.
    Hyprland's natural_scroll only applies to libinput devices, and per-device
    scroll_factor is clamped to 0–100, so there was no way to fix this on the compositor
    side for the virtual pointer.
  - With invert_scroll = true, scroll direction is correct. The local touchpad on the
    receiver is unaffected.
  - Keyboard, pointer motion and buttons keep working normally.
  - The sender still runs the older release and connects fine to this build.
  - cargo test --workspace --locked passes.

  Having this as a receiver-side option is exactly what I needed, since my scroll settings
  on the Mac itself are already correct. Would love to see this merged. Thanks!

Bananoman:
Tested this PR in the opposite direction from the report above, and it solves the problem.

**Setup**
- Sender: Arch Linux (Omarchy), Hyprland 0.56.2, stock `lan-mouse 0.11.0` (Arch package, commit `0d2190e7`)
- Receiver: macOS 26.6.1, arm64, built from this branch (`8e07654`), release profile
- The sender runs the older release and connects fine to this build.

**Problem**

Pointer motion arriving on the macOS side was noticeably accelerated, while motion in the
other direction felt correct. Two OS-level workarounds were ruled out by testing before
trying this PR:

- `defaults write -g com.apple.mouse.scaling -1` on the receiver changed nothing. Events
  injected via `CGEventPost` do not appear to go through the HID acceleration curve.
- Lowering `input.sensitivity` on the sender (`hyprctl eval 'hl.config({input={sensitivity=-0.8,accel_profile="flat"}})'`)
  made the local pointer very slow but did not change what arrived on the Mac. Capture
  happens before the compositor applies its pointer acceleration.

That left receiver-side scaling as the only remaining lever.

**Config on the receiver**

```toml
[input_post_processing]
mouse_sensitivity = 0.5
```

**Result**

Pointer speed on entering macOS now matches the local feel. Keyboard, buttons and scroll
behave normally. `lan-mouse cli set-mouse-sensitivity` is useful for finding the value
without a restart.

Build is clean on macOS arm64: `cargo build --release` with gtk4 4.24.0, libadwaita 1.10.0,
rustc 1.98.1, zero warnings.

Receiver-side is the right place for this. Neither OS exposes a setting that reaches these
events once they are in flight.

Bananoman:
Follow-up to my report above: found one incompatibility worth fixing before merge.

`connect.rs` sends `ProtoEvent::Hello` right after the DTLS handshake, and the comment
says old peers will skip it via the forward-compat handler in `receive_loop`. That
handler is `82766cd`, which is in `main` but not in the tagged v0.11.0. The v0.11.0
receive loop does `log::warn!("error receiving event: {e}"); break;` on any event it
cannot decode, so a v0.11.0 listen side drops the connection the moment the `Hello`
arrives.

Observed with connect side = this branch (macOS) and listen side = stock v0.11.0
(Arch package, commit `0d2190e7`): the pointer could not cross from the Mac to the
Linux box, while the reverse direction worked, since the old peer never sends a
`Hello` and this branch tolerates unknown events. Commenting out the `Hello` send on
my build fixed it; both directions work and the sensitivity scaling keeps working.

Two options, the first is the smaller change:
- State in the PR description and README that both peers need a build that includes
  `82766cd`, since the current tagged release does not.
- Gate the `Hello` send behind a config option that defaults to off until a tagged
  release ships the forward-compat handler.

PR #473, opened 2026-07-27
fix(macos): prevent bidirectional capture loops

mrcha033:
## Problem

On a bidirectional Linux/Wayland ↔ macOS setup, Lan Mouse's macOS event tap
observes the pointer events posted by its own emulation backend. A synthesized
edge event can therefore be mistaken for a new local capture and immediately
send the pointer back, producing an Arch → macOS → Arch capture loop.

## Changes

- tag every macOS event posted by the emulation backend and retain a process-ID
  fallback when CoreGraphics does not preserve the tag
- keep separate physical and emulated edge-arm state
- ignore the initial outward crossing, then re-arm on the first inward motion
  so short reversals work
- send a synthesized return crossing only to the matching `EnterOnly` capture
  while preserving normal default capture for chained topologies
- handle dragged-pointer crossings as motion
- make capture state updates synchronous and release the state lock before
  bounded event delivery
- add routing, short-reversal, drag, and stream-progress regression tests

## Validation

- `cargo clippy -p input-capture --all-targets --all-features --locked -- -D warnings`
- macOS app bundle launches at commit `3d202975`
- recursive `@rpath` dependency closure and `codesign --verify --deep --strict`
  pass locally

Physical Arch → macOS → Arch verification is still running against the Arch
binary from #471 (`83e66bc`), so this is a draft.

## Scope

- the unsuccessful layer-shell re-entry experiment is not included
- macOS dylib packaging changes are not included
- this patch targets one active incoming peer; carrying peer identity through
  CoreGraphics event provenance is left for a follow-up

This is independent of #471, which contains the input post-processing changes.

### DIFF
--- a/input-capture/src/dummy.rs
+++ b/input-capture/src/dummy.rs
@@ -42,6 +42,10 @@ impl Capture for DummyInputCapture {
         Ok(())
     }
 
+    async fn set_enter_only(&mut self, _pos: Position, _enabled: bool) -> Result<(), CaptureError> {
+        Ok(())
+    }
+
     async fn release(&mut self) -> Result<(), CaptureError> {
         Ok(())
     }
--- a/input-capture/src/layer_shell.rs
+++ b/input-capture/src/layer_shell.rs
@@ -628,6 +628,10 @@ impl Capture for LayerShellInputCapture {
         Ok(inner.flush_events()?)
     }
 
+    async fn set_enter_only(&mut self, _pos: Position, _enabled: bool) -> Result<(), CaptureError> {
+        Ok(())
+    }
+
     async fn release(&mut self) -> Result<(), CaptureError> {
         log::debug!("releasing pointer");
         let inner = self.0.get_mut();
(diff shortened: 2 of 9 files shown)


LukePeltier:
Tested this on Arch + niri with macOS as the client. Before this PR, I could move the mouse from Linux to macOS, but fairly often it would get stuck there and I’d have to use the release bind to get back.

With this branch, I’ve been rapidly moving back and forth between Linux and macOS and haven’t been able to reproduce the issue at all so far.

Looks like this fixes it for me. Happy to test anything else if useful.

feschber:
> Tested this on Arch + niri with macOS as the client. Before this PR, I could move the mouse from Linux to macOS, but fairly often it would get stuck there and I’d have to use the release bind to get back.
> 
> With this branch, I’ve been rapidly moving back and forth between Linux and macOS and haven’t been able to reproduce the issue at all so far.
> 
> Looks like this fixes it for me. Happy to test anything else if useful.

Thats nice to hear!

PR #476, opened 2026-07-29
fix(macos): raise inactive windows on synthetic content clicks

### LINKED ISSUE #464: macOS 27: Clicking inactive window content does not focus/raise window
## Environment

- OS: macOS 27 beta
- lan-mouse: tested with v0.11.0
- Setup: controlling the Mac from another machine via lan-mouse

## Problem

After updating to macOS 27, clicking on an inactive window's content area through lan-mouse no longer focuses or raises that window.

The click itself seems to be delivered, but the target window does not become active, so keyboard input does not go to that window. To focus the window, I have to click the title bar / top chrome area first.

## Expected behavior

Clicking anywhere inside an inactive window through lan-mouse should focus/raise that window, matching normal physical mouse behavior.

## Actual behavior

- Clicking window content area through lan-mouse: does not focus/raise the window
- Clicking title bar through lan-mouse: focuses/raises the window
- Using a physical mouse/trackpad directly on the Mac: works normally

## Notes

This looks very similar to a macOS 27 beta issue reported by Deskflow:

https://github.com/deskflow/deskflow/issues/9852

Their report suggests that macOS 27 changed/regressed behavior for synthetic mouse events. A possible workaround mentioned there is to use the macOS Accessibility API to raise/focus the window under the pointer before sending the synthetic mouse click, e.g. using `AXUIElementCopyElementAtPosition` and `kAXRaiseAction`.

So this may be a macOS 27 regression, but lan-mouse may need a compatibility workaround for macOS 27.

mouzhi:
## Summary
- macOS 27 no longer focuses/raises inactive windows when a synthetic click lands on the window body
- title-bar clicks still work, but content-area clicks can interact without activating the target window
- keyboard input then stays on the previous frontmost app

## Fix
Before posting `LeftMouseDown`, look up the Accessibility element under the pointer and:
- `AXRaise` the owning window
- set `AXMain` / `AXFocused` on the window when available
- set `AXFrontmost` on the owning application

This is a compatibility workaround for the macOS 27 synthetic-event regression.

## Test plan
- [x] Build lan-mouse on macOS aarch64
- [x] Control a macOS 27 machine from Windows via lan-mouse
- [x] Click an inactive/background window **title bar** → still focuses/raises
- [x] Click an inactive/background window **content area** → now focuses/raises and accepts keyboard input
- [x] Verify Accessibility + Input Monitoring remain granted after installing the patched binary

Fixes #464

### DIFF
--- a/input-emulation/src/macos.rs
+++ b/input-emulation/src/macos.rs
@@ -10,6 +10,10 @@ use core_graphics::event::{
     ScrollEventUnit,
 };
 use core_graphics::event_source::{CGEventSource, CGEventSourceStateID};
+use core_foundation::base::{CFTypeRef, TCFType};
+use core_foundation::string::CFString;
+use std::ffi::c_void;
+use std::ptr;
 use input_event::{
     BTN_BACK, BTN_FORWARD, BTN_LEFT, BTN_MIDDLE, BTN_RIGHT, Event, KeyboardEvent, PointerEvent,
     scancode,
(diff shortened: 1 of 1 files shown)


feschber:
There is a proper fix for this:
https://github.com/deskflow/deskflow/pull/9978
Let me know if you're willing to implement this.
I'm not going to pull a workaround.

PR #482, opened 2026-08-10
feat: enter a client with a key bind instead of a screen edge

### LINKED ISSUE #260: Feature: Add a keyboard shortcut to change to different screens
Reading around, there are requests for this buried in other requests (https://github.com/feschber/lan-mouse/issues/26#issuecomment-1832877166, https://github.com/feschber/lan-mouse/issues/139#issuecomment-2147995949) - but since I would really like this feature, I'm pulling it out into it's own issue so it doesn't get lost.

Basically, i want what Synergy does (and input-leap, etc): https://symless.com/synergy/help/add-a-hotkey-or-keyboard-shortcut-to-change-to-different-screens

Basically because I tend to do most navigation by keyboard, so switching to mouse just to move my mouse up the page to switch screens is cumbersome for me. Would be great to be able to press my own custom key combo, and have it switch to a different monitor.

khac88:
Closes #260.

Adds optional `[enter_binds]`: key combinations that begin capture at a position
without the pointer having to travel into the corresponding screen edge.

```toml
[enter_binds]
right = [ "KeyLeftCtrl", "KeyLeftAlt", "KeyRight" ]
top = [ "KeyLeftCtrl", "KeyLeftAlt", "KeyUp" ]
```

Motivation from the issue: people who navigate by keyboard don't want to reach for
the mouse just to change machines, edges get crossed by accident, and games that
capture the pointer make the edge unreachable.

## Approach

A bind produces *exactly* the capture an edge crossing produces — the backend calls
the same `start_capture()` and emits the same `CaptureEvent::Begin` — so everything
downstream (`CaptureTask`, service, `enter_hook`, `release_bind`) works unchanged and
there is no second notion of "entered".

Consequences that fall out of that, rather than being special-cased:

- a bind only fires while capture is inactive; `release_bind` is still what returns
  control to the local machine
- a bind only fires for a position that currently has an active client, so it can
  never enter a client an edge crossing could not

Binds are keyed by **position**, not by client, for the same reason: crossing an edge
enters every client at that edge, so a per-client bind would have been a per-position
bind wearing a disguise — with the added problem that the frontend's `save_config`
round-trip has no representation for it and would have had to reconstruct it.

The matching itself lives in one small, backend-independent type
(`input-capture/src/enter_bind.rs`) that both backends drive, with unit tests
covering: firing on the last key down, key order independence, never firing on
release, ignoring positions without a client, extra held keys not blocking the bind,
and the two ways stale held-keys could otherwise fire a bind (below).

`Capture::set_enter_binds` has a default no-op implementation, so the Linux backends
are untouched and keep entering by edge only — the input capture portal doesn't
deliver input while capture is inactive, as noted in #260.

## Notes for review

- Held keys are forgotten whenever capture begins, by *either* route. Keys stop being
  observed for the duration of a capture, so without this a key held when capture
  began (and released during it, unseen) would still count as held afterwards, and
  could complete a bind on its own later.
- The bind chord is consumed locally and not forwarded, so the remote starts with no
  keys held. That mirrors `release_capture`, which deliberately releases everything
  before handing control back — clean slate in both directions. Carrying held
  modifiers across the switch instead would be a separate change.
- On macOS the idle branch decodes keys through the existing `get_events()` rather
  than reading the `CGEvent` separately, so modifier bookkeeping stays continuous
  across the idle → capturing transition instead of a modifier held at capture start
  being re-reported as a fresh press.
- A key event carries no meaningful pointer location, so the macOS path queries the
  window server for the real cursor position instead of reading it off the triggering
  event; `start_capture` now takes a `CGPoint` for that reason. Windows uses
  `GetCursorPos` and seeds both `ENTRY_POINT` and `PREV_POS` so the first motion
  delta after entering is ~0 rather than a jump.

## Testing

- `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -D warnings`
  and `cargo test --workspace --all-features` are clean on macOS, and `cargo check` /
  `cargo clippy` are clean for `x86_64-pc-windows-msvc`.
- macOS verified end-to-end against a Windows peer over Tailscale: pressing the bind
  from the idle state enters the client exactly as an edge crossing does —

  ```
  input_capture::macos] entering client @ top: enter-bind pressed
  lan_mouse::service]   entering client 0 ...
  lan_mouse::connect]   client (0) connected @ …:4242
  lan_mouse::capture]   client 0 acknowledged the connection!
  lan_mouse::capture]   releasing capture: left remote client device region
  ```

  Confirmed on real hardware too: pressing the bind physically on the Mac hands input
  to the Windows machine — moving the trackpad afterwards moves the remote pointer.
  Entering by edge afterwards is unaffected and `release_bind` still hands control
  back.

  Note the pointer lands wherever it was left on the remote rather than at a matching
  position, since `Enter` carries no coordinates — same as an edge crossing, and
  orthogonal to this PR (cf. the `CursorPos` proposal in #429).
- Windows: implemented and compile/clippy-checked for `x86_64-pc-windows-msvc`, but
  not exercised at runtime — the hook path mirrors the macOS one and shares the tested
  matcher, so a confirmation from someone running the Windows capture backend would be
  welcome.

## Possible follow-ups (deliberately not in this PR)

- switching straight from one client to another while capture is already active
  (today: release, then bind)
- an option to disable edge crossings entirely and rely on binds only, which a couple
  of people asked for in #260
- entering at a position matching where the pointer was — that needs coordinates on
  the wire, which is its own discussion (#429), and binds inherit whatever edge
  crossings end up doing

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,6 +385,12 @@ release_bind = [ "KeyA", "KeyS", "KeyD", "KeyF" ]
 # optional port (defaults to 4242)
 port = 4242
 
+# optional key binds that enter the client(s) at a position without
+# the pointer having to cross that screen edge (see "Entering a client
+# with a key bind" below)
+[enter_binds]
+right = [ "KeyLeftCtrl", "KeyLeftAlt", "KeyRight" ]
+
 # list of authorized tls certificate fingerprints that
 # are accepted for incoming traffic
 [authorized_fingerprints]
@@ -414,6 +420,39 @@ port = 4242
 ```
 
 Where `left` can be either `left`, `right`, `top` or `bottom`.
+
+### Entering a client with a key bind
+
+An `[enter_binds]` entry makes a position reachable by holding a
+combination of keys, so the pointer no longer has to be moved into the
+corresponding screen edge:
+
+```toml
+[enter_binds]
+right = [ "KeyLeftCtrl", "KeyLeftAlt", "KeyRight" ]
+top = [ "KeyLeftCtrl", "KeyLeftAlt", "KeyUp" ]
+```
+
+Binds are keyed by position rather than by client because entering is
+position-based: crossing an edge enters *every* client at that edge,
+and a bind is deliberately no different.
+
+The bind takes effect on the machine that *sends* input and behaves
+exactly like the matching edge crossing: capture begins, the pointer is
+warped to that edge and the usual `release_bind` returns control to the
+local machine. Consequently a bind only fires while capture is
+inactive, and only for a position that currently has an active client.
+
+The key combination itself is consumed locally and is not forwarded, so
+the remote machine starts with no keys held — mirroring `release_bind`,
+which releases everything before handing control back. Keep holding the
+bind's modifiers after switching and the remote will not see them as
+held; release and press them again for that.
+
+> [!NOTE]
+> Supported on the macOS and Windows capture backends. The Linux
+> backends observe no input while capture is inactive, so binds are
+> ignored there (see [#260](https://github.com/feschber/lan-mouse/issues/260)).
 
 ## Roadmap
 - [x] Graphical frontend (gtk + libadwaita)
--- a/input-capture/src/enter_bind.rs
+++ b/input-capture/src/enter_bind.rs
@@ -1,0 +1,231 @@
+use std::collections::{HashMap, HashSet};
+
+use input_event::scancode;
+
+use crate::Position;
+
+/// Tracks the keys held down while capture is inactive and reports
+/// when the bind configured for a position becomes fully held.
+///
+/// Backends that can observe key events outside of an active capture
+/// (a global keyboard hook or event tap) feed every such key into
+/// [`Self::key_event`] and begin capture at the returned position,
+/// exactly as they would for a pointer crossing the screen edge.
+#[derive(Debug, Default)]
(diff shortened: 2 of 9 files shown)


feschber:
The main issue with whole enter bind is that it's not compatible with wayland in the way you are implementing it here.

We can still talk about implementing this for the systems that do support this currently (X11, windows, macos), but I'd like it to be in a way that is compatible API wise with wayland in the future.

That means:
`input-capture` _can not_ have a `set_enter_bind` function in its public interface.
`input-capture` as a crate right now sets up barriers and starts capturing when a barrier is entered.

What we _can_ do is add another public interface `activate(pos)` or preferably `activate_handle(capture_handle)`.
I'd be in favor of the second option, even though both probably have their uses.

The trigger signal for these functions has to come from elsewhere, something outside of the `input-capture` crate.
On wayland this would be the global-shortcuts desktop portal or the new `xx-hotkey-v1: global hotkey protocol`
 wayland protocol. 
I'm not familiar with windows but a google search brought up the `RegisterHotKey` function that takes a callback and I think there is something similar on MacOS as well.

In its current state you have a lot of duplicate code in the platform specific backends in a place where its not compatible with all desktops.
If you want to continue with this, I'd be very happy if you want to rewrite this in the way I described.

PR #485, opened 2026-08-20
fix(input-capture/layer_shell): Call read() on guard only once

ids1024:
With `wayland-backend/client_system` enabled, this was causing an infinite loop since `read` never returned `WouldBlock`.

This seems to be an inconsistency in wayland-rs: https://github.com/Smithay/wayland-rs/pull/959

In any case, it isn't necessary to call `read` more than once.

### DIFF
--- a/input-capture/src/layer_shell.rs
+++ b/input-capture/src/layer_shell.rs
@@ -123,7 +123,6 @@ struct State {
     focused: Option<Arc<Window>>,
     global_list: GlobalList,
     globals: Globals,
-    wayland_fd: RawFd,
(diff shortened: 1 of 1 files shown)


feschber:
Just for my understanding: It always returned 0?

ids1024:
https://docs.rs/wayland-client/latest/wayland_client/backend/struct.ReadEventsGuard.html#method.read returns "the number of dispatched events", like `wl_display_dispatch_queue_pending` in the libwayland C library.

Though to be honest, I don't know whether or not there's anything this return value is actually useful for.

PR #487, opened 2026-08-28
fix: refresh macOS display bounds on hotplug

grazzolini:
## Summary

- Recompute macOS capture bounds from the current active-display union after monitor changes.
- Refresh bounds during pointer motion when Quartz callbacks are missed or arrive before topology settles.
- Reanchor an active capture to the updated edge while preserving and clamping the cross-axis position.

## Testing

- `cargo fmt --check`
- `cargo test -p input-capture`
- `cargo clippy -p input-capture --all-targets`
- Manually switched the left external display from macOS to Linux and crossed from the new macOS left edge without restarting Lan Mouse.

### DIFF
--- a/input-capture/src/macos.rs
+++ b/input-capture/src/macos.rs
@@ -31,6 +31,7 @@ use std::{
     sync::{Arc, OnceLock},
     task::{Context, Poll, ready},
     thread::{self},
+    time::{Duration, Instant},
 };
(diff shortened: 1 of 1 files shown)


grazzolini:
I have been using this patch for a couple of months and it works. This has been tested with 3 monitors and many combinations among them, including changing the geometry a lot.

Issue #26, opened 2023-09-20
[Feature Request] Toggle Bind to switch between clients


Niahex:
Thank you for your work!

I have a suggestion:

- Set up a key to lock entry into the client.

Thanks.

CupricReki:
To expand on this, it would be nice to have keyboard shortcuts for switching between clients.

Ph-ill:
Just wanted to add a +1 to this request. I'm running a two-machine setup with lan-mouse on Arch Linux/Hyprland and a keyboard shortcut to move focus between clients would be a really natural addition to the CLI. The current edge-of-screen behaviour works well but a hotkey option would make the workflow even smoother. No rush, just wanted to show there's ongoing interest in this feature!

ameyakakade:
There could also be command line options for the same, to make it easy to integrate with existing keybinds.

PR #493, opened 2026-09-02
feat(capture): allow restricting input to host

TibiIius:
I came from Barrier to this project since Barrier is no longer maintained and does not work on the Wayland compositor I use. So far I'm very happy with my switch! :)
One feature I'm currently missing however is having a key bind to allow restricting all input to the server/host system on-demand. I gave it my best effort trying to implement this feature here. One drawback currently is that modifier keys do not sync to their state but rather just toggle the restriction on/off, as I've not been able to find a way to easily read the state of a modifier. This means that if you'd start Lan Mouse with scroll lock engaged (assuming scroll lock is the bind used), the toggle logic is reversed, i.e. scroll lock on -> restriction off, scroll lock off -> restriction on.

I'm marking this as a draft for now as I have yet to actually test it. I mainly created the PR in such an early state to avoid wasted/redundant effort in case someone else wants to implement this feature as well.

Disclaimer: I used a coding agent mainly to get an overview about what I can work with more quickly (structs, pub functions, etc) and to refactor. I have manually verified all changes made. The edge-trigger logic was suggested by AI, so I am unsure as to whether this is actually needed (i.e. if holding down the key bind would otherwise toggle the state repeatedly).

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,11 +376,19 @@ To create this file you can copy the following example config:
 > to their names in [input-event/src/scancode.rs#L172](input-event/src/scancode.rs#L176).
 > This is bound to change
 
+> [!TIP]
+> Press the jail bind (*ScrollLock* by default) to toggle the "mouse jail".
+> While enabled, your mouse stays confined on your host and cannot cross a
+> screen edge onto a connected peer — press the bind again to disable.
+
 ```toml
 # example configuration
 
 # configure release bind
 release_bind = [ "KeyA", "KeyS", "KeyD", "KeyF" ]
+
+# configure jail bind
+jail_bind = [ "KeyScrollLock" ]
 
 # optional port (defaults to 4242)
 port = 4242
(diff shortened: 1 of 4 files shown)


feschber:
By host / server, I'm assuming you are referring to the device with the keyboard connected?
One thing to keep in mind, which I'm not sure you aware of: Input Capture does not globaly listen to input events, since thats incompatible with wayland.
It only activates once the cursor crosses a "barrier".
So to actually trigger the jail without currently controlling a remote client, you'd have to use the [global shortcuts portal](https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.GlobalShortcuts.html) or the new
[xx-hotkey-v1 protocol](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/525) in wayland.

#26 #260 #482 also need this.

We should probably add a central place first to allow registering hotkeys for all these kinds of features.

feschber:
And FYI: [deskflow](https://github.com/deskflow/deskflow) is the maintained fork of barrier and is actively being developed.

PR #495, opened 2026-09-04
nix(chore): set nixpkgs to nixos-26.05, update flake.lock

### LINKED ISSUE #492: [Nix flake] The flake uses crates.io/api to fetch dependencies, resulting in a 403 error
Recently it seems like the Crates.io API has clamped down on the use of their API for fetching packages. It seems now the recommendation is to use static.crates.io. See this thread for details:

https://discourse.nixos.org/t/403-error-when-fetching-packages-from-crates-io/79803/21

I've had to disable the flake in order to update my computer today unfortunately.

onelocked:
Closes #492

Switched the `nixpkgs` input to `nixos-26.05`. I chose this version because it's the latest stable release and the last one that supports `x86_64-darwin` (unfortunately, 26.11 sunsets support for `x86_64-darwin` builds).

Bumped the flake inputs using `nix flake update`

I ran the build locally on my x86_64-linux machine it complied successfully and it works.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,16 +2,16 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1772963539,
-        "narHash": "sha256-9jVDGZnvCckTGdYT53d/EfznygLskyLQXYwJLKMPsZs=",
+        "lastModified": 1788405554,
+        "narHash": "sha256-r2f1oUwixlgq9zOdYLqJLfS/lWBT60/IITjhTKI59JU=",
         "owner": "nixos",
         "repo": "nixpkgs",
-        "rev": "9dcb002ca1690658be4a04645215baea8b95f31d",
+        "rev": "a5cc6f2c37bf518436dc8d1c288ccd0c43c2f4c4",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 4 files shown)


onelocked:
Fixed the build error on MacOS (aarch64-darwin), it was failing due to appstream putting literal string 'none required' into linker flags, this issue is not present on `nixos-unstable` branch, the easy solution would've been to switch to nixos-unstable but I did not want to break builds for `x86_64-darwin`, so I made a small workaround and commented it on the flake.

Once NixOS 26.05 is deprecated, I will return to bump the flake, remove the appstream workaround, and drop x86_64-darwin from the Nix builds. Until then I wanted to keep supporting the `x86_64-darwin` platform as much as possible.

Additionally there was a naming change
`evaluation warning: stdenv.isLinux is deprecated, use stdenv.hostPlatform.isLinux instead`
`evaluation warning: stdenv.isDarwin is deprecated, use stdenv.hostPlatform.isDarwin instead`
So I went ahead and replaced those instances. 


Huge thanks to @KristijanZic for helping me debug and testing the build on their Mac!

feschber:
Thank you!

Issue #318, opened 2025-08-09
slow scrolling chrome with emulation=wlroots capture=layer-shell


selckin:
Hello,

Using niri as compositor on both sides resulting in: emulation=wlroots capture=layer-shell 

Mouse scrolling works fine in terminals, but only scrolls very small amount in google-chrome (in wayland mode)

Using 'wev' to show events, using the real mouse shows
```
[        15:      wl_pointer] axis_source: 0 (wheel)
[        15:      wl_pointer] axis_value120: axis: 0 (vertical), value120: 120
[        15:      wl_pointer] axis_relative_direction: axis: 0 (vertical), direction: 0
[        15:      wl_pointer] axis: time: 50410752; axis: 0 (vertical), value: 15.000000
```
Using the lan-mouse shows:
```
[        15:      wl_pointer] axis_source: 2 (continuous)
[        15:      wl_pointer] axis_value120: axis: 0 (vertical), value120: 1
[        15:      wl_pointer] axis_relative_direction: axis: 0 (vertical), direction: 0
[        15:      wl_pointer] axis: time: -1913142096; axis: 0 (vertical), value: 20.000000
```

Patching lan-mouse to pass the original value results in scrolling working for me in both terminals & google-chrome:

```
diff --git i/input-emulation/src/wlroots.rs w/input-emulation/src/wlroots.rs
index a36c7a1fc330..d860c0aebfd8 100644
--- i/input-emulation/src/wlroots.rs
+++ w/input-emulation/src/wlroots.rs
@@ -208,11 +208,11 @@ impl VirtualInput {
                         self.pointer.frame();
                     }
                     PointerEvent::AxisDiscrete120 { axis, value } => {
                         let axis: Axis = (axis as u32).try_into()?;
                         self.pointer
-                            .axis_discrete(now, axis, value as f64 / 6., value / 120);
+                            .axis_discrete(now, axis, value as f64 / 6., value);
                         self.pointer.frame();
                     }
                 }
                 self.pointer.frame();

```

Regards

selckin:
Also need Wheel axis source, else scrolling over chrome (pinned) tabs will skip every other one

```patch
diff --git a/input-emulation/src/wlroots.rs b/input-emulation/src/wlroots.rs
index a36c7a1fc330..f5960425a5bd 100644
--- a/input-emulation/src/wlroots.rs
+++ b/input-emulation/src/wlroots.rs
[...]

Issue #133, opened 2024-05-15
Mouse wheel distance/speed


xf-:
Still broken - build the client and server on latest main commit

Server & client Wayland to Wayland

Fixed - inverted scroll direction

feschber:
What desktop? KDE?

xf-:
> What desktop? KDE?

yes on both

feschber:
I've tried pretty much anything but unfortunately there seems to be a bug with the KDE remote desktop portal implementation.

It might just be that they report 0 discrete scroll steps here: https://invent.kde.org/plasma/kwin/-/blame/master/src/backends/fakeinput/fakeinputbackend.cpp#L153

Basically there is two kinds of scroll events: Axis and AxisDiscrete120.
Axis is for touchpads and given in pixels, while AxisDiscrete120 is given in wheelsteps with each step being equal to 120 in value.

Now the xdg-desktop-portal-kde, which is being used here calls to the `fake_input` protocol: 
https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/blob/master/src/remotedesktop.cpp?ref_type=heads#L322
https://invent.kde.org/plasma/xdg-desktop-portal-kde/-/blob/master/src/waylandintegration.cpp?ref_type=heads#L414
while translating discrete scroll events into fakeinput "Axis" events which are handled in kwin:

https://invent.kde.org/plasma/kwin/-/blame/master/src/backends/fakeinput/fakeinputbackend.cpp#L132

Kwin calls to `pointerAxisChanged` while setting discrete Scrolling steps to 0...

So all of this seems to be fundamentally broken in KWIN itself and I fear there is nothing I can do right now to fix it.

I will however make sure to report it upstream

xf-:
THX for your work. I still have my patched build and it works.

feschber:
> THX for your work. I still have my patched build and it works.

It's not perfect though, right? In firefox, at least from my testing it always jumped around a bit and did not scroll smoothly (instead jumping to the next possition directly).
Regardless, I can see if I can replicate the behaviour to what your patch looks like as closely as possible.

Do you mean the state before e21ab02a6e752d39ddcf73ceb9c8e5e50d014923 or the current master with e21ab02a6e752d39ddcf73ceb9c8e5e50d014923 reverted?

xf-:
> > THX for your work. I still have my patched build and it works.
> 
[...]

Issue #499, opened 2026-09-16
Proposal: opt-in batched binary wire format (99.5% less wire traffic, one frame() per datagram)


cawa0505:
Hi, and thanks for maintaining lan-mouse — great project. I've been running it as the main KVM between my desktops and experimented with an **opt-in batched binary wire format** for input events. Before opening any PR, I'd like to gauge whether interest exists upstream, since this touches the wire format.

## Motivation

At high polling rates, the current protocol sends one datagram per input event (each event wrapped into the legacy frame, e.g. 21 bytes for motion). On the receiving side each datagram becomes an individual `wl_pointer` event stream with its own `frame()`. I wanted to try batching adjacent events into a single datagram to reduce packet rate and per-event overhead in the emulation backends (one `frame()` per datagram).

## What the prototype does

- **4-byte header**: magic `0x4C` (doubles as format version), `event_count: u8` (max 64), `seq: u16` per peer for stale/reordered datagram dropping
- **Compact payloads**: motion 3 B (dx/dy i8, falling back to a 5 B i16 variant), button/key/wheel 3 B each (wheel in 1/120 discrete ticks)
- **Motion coalescing**: adjacent motion accumulates in an i32 accumulator and is emitted once per flush — one `frame()` per datagram
- **Flush policy**: batch full (64), any non-motion event, 4 ms timer, capture release
- **Fully opt-in**: gated behind a `batched_protocol` config flag (default on in my fork, off = byte-identical legacy behavior). The magic byte `0x4C` is an invalid event id in the legacy format, so both directions can multiplex cleanly; a legacy peer that receives a batched datagram currently IGNOREs it via the invalid-event-id path (silent), and the existing `Hello` commit-hash soft-warn would surface the mismatch.
- Linux backends only so far (verified end-to-end on niri + a wlroots compositor)

## Measured (synthetic round-trip, release build)

10 000 motion events (~10 s of 1 kHz movement), coalesced:

| | Bytes | Datagrams |
|---|---|---|
| Legacy encoding | 210 000 B | 10 000 |
| Batched encoding | 1 099 B | 157 |

Wire size reduction: **~99.5 %**. CPU cost is negligible: encode ≈ 18 ns/event, decode ≈ 15–22 ns/event (Zen 3). The numbers are asserted by a unit test and re-measurable via `cargo test -p lan-mouse-proto --release -- --ignored --nocapture bench` in my fork.

## Links

[...]

PR #500, opened 2026-09-17
fix(emulation): accumulate high-resolution scroll instead of discarding it

wyrickre:
## Problem

On a machine with a high-resolution scroll wheel, scrolling to a remote client does nothing at all, while pointer motion and clicks work perfectly.

High-resolution wheels report *fractions* of a detent rather than whole 120-unit steps. A Logitech MX Master 3 emits values of **16, 24, 32, 64, 72, 80** — never 120. Captured with `lan-mouse test-capture` (layer-shell backend, COSMIC):

```
position: 0 (left), event: scroll-120 (0, -24)
position: 0 (left), event: scroll-120 (0, -16)
position: 0 (left), event: scroll-120 (0, 64)
```

Two emulation backends assume whole steps and discard all of them:

- **`xdg_desktop_portal`** — `value / 120` is integer division, so any `|value| < 120` becomes `0` and `notify_pointer_axis_discrete` is called with no steps.
- **`libei`** — the raw value120 goes to `scroll_discrete()`, and a receiver that only acts on whole steps (mutter) drops it.

The telling symptom is that a *hard flick* scrolls, because it happens to clear 120 within a single event, while ordinary scrolling does nothing. That asymmetry is what distinguishes this from "the scroll device was never bound", which presents identically.

`wlroots` is unaffected because it also sends the continuous component (`value as f64 / 8.`) — likely why #318/#325 describe merely *slow* scrolling there rather than none.

## Fix

A small `Scroll120Accumulator` that sums deltas per axis and emits whole 120-unit steps, retaining the sub-step remainder. Used by both affected backends.

Truncation is toward zero, so the remainder keeps the sign of the accumulated scroll and a direction change cancels rather than latching.

**Classic wheels are unaffected**: 120 in still yields exactly one step out.

## Testing

Five unit tests, including a replay of the real 24-per-tick MX Master case. `cargo fmt --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings` and `cargo test` are all clean.

Verified end to end on **GNOME 46 / Wayland with the `libei` backend** (capture side was COSMIC/layer-shell): scrolling was completely dead before this change and behaves like a normal wheel after.

**The `xdg_desktop_portal` change is by inspection only** — same root cause and more obviously broken (`value / 120` → `0`), but I have not exercised that backend at runtime. Happy to drop that hunk if you would rather it came with its own testing.

## Not included

A configurable scroll step. A threshold knob would control *speed* rather than fineness, since whole steps are always emitted; genuinely fine-grained scrolling would mean emitting the continuous component the way `wlroots` does, which risks double-counting on receivers that honour both and deserves its own change. #133 and #471 already cover that ground.


feschber:
I think it would make more sense to accumulate on the capture side but ultimately we need both anyway

wyrickre:
I'd argue the emulation side is actually the right place for this, and that capture-side accumulation would be a regression.

**Accumulating at capture is lossy, and `wlroots` already depends on the information it would discard.** If capture only forwards whole 120 steps, the sub-detent resolution never reaches the receiver. `wlroots` emulation does:

```rust
self.pointer.axis_discrete(now, axis, value as f64 / 8., value / 120);
```

— raw value for the *continuous* component, quotient for the discrete one. That is exactly why wlroots degrades to merely *slow* scrolling instead of nothing, and smooth scrolling is what a hi-res wheel user wants. Pre-accumulating upstream of it would throw that away and make wlroots worse than it is today.

The sender also cannot know what the receiver will do with the extra resolution: Windows and macOS sinks want whole steps, `wlroots` wants the fractions. Keeping the full-resolution value on the wire and letting each emulation backend decide is the split `wl_pointer` itself uses (`axis_value120` alongside `axis`), and it is the one that does not lose data.

Capture-side accumulation is a reasonable *optimisation* for event-rate and wire traffic (#499 is after the same thing), but it is not a substitute for this: even with it, `libei` and `xdg_desktop_portal` would still drop any remainder they were handed. This fix is needed either way.

As it stands this is a self-contained correctness fix — scrolling goes from completely dead to working on GNOME 46/Wayland via `libei`, no behaviour change for classic wheels (120 in, one step out), five unit tests, and fmt/clippy/test all clean. I would rather not grow its scope; the capture-side change is a separate concern and better as its own PR by whoever wants that optimisation.

